PR #36, opened 2026-04-07
feat: Implement LLM-as-a-Judge for summarization quality

PaulvH63:
Changes:
- Add quality_score attribute to response
- quality_score = 0.6 * faithfulness (check hallucinations) + 0.3 * coverage (is summary complete?) + 0.1 * clarity (writing style)
- Add some tests
- Improve summarization prompt

### DIFF
--- a/src/qfa/services/orchestrator.py
+++ b/src/qfa/services/orchestrator.py
@@ -46,7 +46,10 @@ _SYSTEM_MESSAGE_TEMPLATE = (
 _DEFAULT_SUMMARIZATION_PROMPT = (
     "Summarize the feedback item as concise bullet points.\n"
     "Also create a short descriptive title.\n"
-    'Return valid JSON with exactly these fields: {"title": "...", "summary": "- point 1\\n- point 2"}.\n'
+    "Assign quality_score: a float from 0.0 to 1.0 for how well the summary "
+    "captures the feedback (1.0 = fully faithful and complete).\n"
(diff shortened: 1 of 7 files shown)


src/qfa/services/orchestrator.py
```diff
@@ -61,6 +107,27 @@
...
+    except ValueError as exc:
+        raise AnalysisError("LLM judge returned invalid quality score") from exc
+    if not 0.0 <= score <= 1.0:
+        raise AnalysisError("LLM judge returned quality score outside 0.0-1.0")
+    return score
+
+
+def _build_judge_system_message(source_text: str, summary: str) -> str:
+    """Fill the judge prompt without str.format (source may contain ``{``)."""
+    return _JUDGE_PROMPT.replace("{source_text}", source_text).replace(
```
mariushelf: this works, but only if `source_text` does not contain "{summary}". Since you are using the curly braces in `_JUDGE_PROMPT` you could use `str.format` --> https://docs.python.org/3/library/stdtypes.html#str.format

i.e. `_JUDGE_PROMPT.format(source_text=source_text, summary=summary)`

mariushelf (changes requested):

src/qfa/services/orchestrator.py
PaulvH63: Good catch! 
Fixed.

PR #47, opened 2026-04-16
feat: release version included in the api docs, instead of hard-coded value

teeuwski:
Reuse existing `qfa.__version__` functionality for showing the correct version tag in the fastapi app


mariushelf (approved):
LGTM!

PR #48, opened 2026-04-16
feat: auto-update semantic version in lock file

teeuwski:
## Problem

The current `semantic-release` implementation only increases the releases in the pyproject.toml, but does not  update the version in the `uv.lock` file.

## Solution

This functionality from [the documentation of `python-semantic-release`](https://python-semantic-release.readthedocs.io/en/latest/configuration/configuration-guides/uv_integration.html#updating-the-uv-lock) adds this functionality.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -36,6 +36,9 @@ dependencies = [
 Homepage = "https://github.com/rodekruis/qualitative-feedback-analysis"
 Repository = "https://github.com/rodekruis/qualitative-feedback-analysis"
 
+[project.optional-dependencies]
+build = ["uv ~= 0.11.7"] # Included for the build command in semantic-release
+
 [tool.hatch.build]
 packages = ["src/qfa"]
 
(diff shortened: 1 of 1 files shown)


teeuwski:
@copilot can you somehow test if this would work? I would prefer not to test on the main branch. Place your findings here, and propose solutions that are minimal if this does not work. Read the docs that I linked to in this PR if you're looking for solutions.

mariushelf:
This looks good in terms of code, but which problem does it solve? Or is it just adding a best practice to prevent problems in the future?

teeuwski:
@mariushelf if I run anything using uv run after the semantic update, it will update my uv.lock file first with the new version.

e.g.:
image

mariushelf (approved):
Ah ok, so that happens locally after you pull after a release, right?

LGTM!

teeuwski:
Yes exactly! I'll merge

PR #55, opened 2026-04-21
Feat/ubiquitous language

mariushelf:
Ubiquitous language as defined by @DGorsseRedCross in #23 .

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,6 +14,11 @@ Each request contains dozens to thousands of feedback documents.
 The documents need to be analysed and, and the result sent back to the CRM system in a
 synchronous API call.
 
+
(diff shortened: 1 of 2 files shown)


teeuwski:
Does the current state follow these guidelines? Could be interesting to include in this PR if not

mariushelf:
> Does the current state follow these guidelines? Could be interesting to include in this PR if not

not all of it. some of the implementation predates the document, some of it ignored it because it was hidden in a closed github issue.

I'd prefer to agree on this document first and merge, and then fix the codebase incrementally/separately.

DGorsseRedCross (approved):
Have made changes and looks good now.

PR #56, opened 2026-04-21
feat: Feedback item coding by LLM

PaulvH63:


teeuwski (approved):
lgtm🚀

I think this would benifit a lot from a refactoring that I'm working on in [this branch](https://github.com/rodekruis/qualitative-feedback-analysis/tree/refactor/structured-llm-calling). I'll rebase that off the newly created main and rework.

PR #57, opened 2026-04-21
feat(pii): implement text anonymization using Presidio

### LINKED ISSUE #30: Investigate pre-processing models to anonymize (remove PII)
### **Who is your contact person for this task?**
A name of the person this work should be handed over to, can be internal. 

### **User Story**
As a [end-user]
I want to [action/functionality]
So that I [reason/benefit]

### **What needs to be done**
- [ ] Task 1
- [ ] Task 2

### **Acceptance Criteria**
- [ ] Example
- [ ] Tested and reviewed
- [ ] Documented the changes in the approriate location
- [ ] Handed over to requester

teeuwski:
Implements the [Presidio](https://microsoft.github.io/presidio/) package using SpaCy for anonymizing PII. 

This implementation auto-downloads a [394 MB Spacy model](https://spacy.io/models/en#en_core_web_lg) that is used to identify and obscure this information.

Any call routing through `aysync def _call_with_retries` will anonymize the `user_message` by default. It will also de-anonimize the call after.

Note that this anonymizes the following information (credits to Gemini):

| Entity Type   | Description                                               |
|---------------|-----------------------------------------------------------|
| PERSON        | Full names, first names, and last names.                  |
| EMAIL_ADDRESS | Standard email formats (e.g., user@example.com).          |
| PHONE_NUMBER  | Telephone numbers (validated via logic and patterns).     |
| LOCATION      | Cities, countries, provinces, and geographical landmarks. |
| DATE_TIME     | Dates, times, or relative time periods.                   |
| IP_ADDRESS    | Both IPv4 and IPv6 addresses.                             |
| URL           | Web links and uniform resource locators.                  |
| CREDIT_CARD   | Card numbers (validated via the Luhn checksum).           |
| CRYPTO        | Crypto wallet addresses (primarily Bitcoin currently).    |
| IBAN_CODE     | International Bank Account Numbers.                       |
| MAC_ADDRESS   | Media Access Control addresses for network hardware.      |
| NRP           | Nationality, Religious, or Political group affiliations.  |

An important note is that the spacy model is a local, which means that we should monitor the performance on higher loads. 

Closes #30


teeuwski:
I'm currenlty expanding on this as I'd like to use substitution logic from the Presidio package allows for the output of the endpoint to contain the un-redacted inputs names.

DGorsseRedCross (changes requested):
Thanks @teeuwski! Highly relevant from a data protection and governance POV.

Comments from my side before we implement:
1. Regarding anonymizing LOCATION data: we don't need/want specific addresses of persons indeed, but we do need higher level location data to be included (village, city, region, province, etc.). How do we ensure this is kept for analysis? 
2. DATE_TIME does not necessarily hold PII and is in many feedback records relevant context. I would suggest not to include this.
3. AFAIK, we never store IP_ADDRESS, CRYPTO and MAC_ADDRESS but can not harm to include this, unless you think otherwise (for computing/load reasons or similar).
4. Can we disable this for synthetic/research datasets? 
5. Small suggestion: it could be useful if users can detect whether an output was generated on anonymised input (e.g. metadata flag), purely for interpretability and auditing purposes.

teeuwski:
> 1. Regarding anonymizing LOCATION data: we don't need/want specific addresses of persons indeed, but we do need higher level location data to be included (village, city, region, province, etc.). How do we ensure this is kept for analysis?

The LLM never receives these names, however we de-anonymize after. This means that the semantic structure is kept. I could, however see that there could be a relationship between Amsterdam and Netherlands for instance that is lost if it sees LOCATION_0 and LOCATION_1 instead. For now, I've kept it- around, but this can be removed from the anonymization. Note that with this implementation, it appears to be a binary options to include all or no location info directly.

> 2. DATE_TIME does not necessarily hold PII and is in many feedback records relevant context. I would suggest not to include this.

Done.  d6de3255b0eb70616c065641c1cc68805fd9b706

> 3. AFAIK, we never store IP_ADDRESS, CRYPTO and MAC_ADDRESS but can not harm to include this, unless you think otherwise (for computing/load reasons or similar).

I'm going to leave that in for now then.

> 4. Can we disable this for synthetic/research datasets?

https://github.com/rodekruis/qualitative-feedback-analysis/blob/7c19f413c266ed3fadde127736dd2d35ffff2168/src/qfa/services/orchestrator.py#L467-L474

This boolean operator allows us to specify if we want to anonymize before sending to the LLM. For testing, I'd prefer using it WITH the anonymization as this gives a good idea of the overall peformance of the system at runtime.

> 5. Small suggestion: it could be useful if users can detect whether an output was generated on anonymised input (e.g. metadata flag), purely for interpretability and auditing purposes.

Users can now give a  `deactivate_anonymization` option in their requests, and this is also included in a response. See bcd290be829fdd7eb98d361666aa76eca940f314

DGorsseRedCross (approved):

PR #60, opened 2026-04-22
feat(sensitive): create endpoint for sensitive document detection

### LINKED ISSUE #18: feat: detect sensitive documents
We need to be able to flag sensitive documents, such as very personal topics, corruption, etc that need to be handled with extra care.

If in doubt, the model should have a tendency to flag documents as sensitive.

## Input
```
{
documents: list[Document],
}
```

with
```
Document: 
  id: str
  content: str
```

## Output:

Sensitivity flags.

```
{
sensitive: list[
   {
      id: str,  # input document id
      is_sensitive: bool
      explanation: str  # why does the model think it's sensitive?
  }
]
}
```

## Open questions
* can we return a confidence score per document?
* how exactly are sensitive topics defined?

teeuwski:
Closes #18 

Also outputs given document ids, so will use the fix of #106 to make sure we don't make the same mistake twice.


teeuwski:
@copilot please implement the changes requested above. I am unsure if the changes to the Enum will still parse the model (inc. description) to the LLM. Please validate this with their internal docs.

src/qfa/api/schemas.py
```diff
@@ -328,6 +328,62 @@ def verify_all_codes_have_same_depth(self) -> "ApiCodingLevels":
...
+    ----------
+        feedback_items : list[ApiFeedbackRecordInput]
+    """
+
+    feedback_items: list[ApiFeedbackRecordInput] = Field(
+        min_length=1,
+        description="List of feedback items to check for sensitive content.",
+    )
+
+    anonymize: bool = Field(
```
mariushelf, line 344: should we even allow users to override this? We have flagged it in #76 already. Are there any use cases that we foresee where users should disable anonymization?

We can also defer this decision until later and merge as is.

src/qfa/api/schemas.py
```diff
@@ -328,6 +328,62 @@ def verify_all_codes_have_same_depth(self) -> "ApiCodingLevels":
...
+        Explanation for the sensitivity rating.
+    sensitivity_types : list[str]
+        Sensitivity categories detected for the feedback item.
+    """
+
+    id: str = Field(description="Identifier of the source feedback item.")
+    is_sensitive: bool = Field(
+        description="Indicates whether the feedback item is considered sensitive."
+    )
+    explanation: str = Field(description="Explanation for the sensitivity rating.")
```
mariushelf, line 369: do we also need a confidence score here? Or is that not necessary at this stage?

mariushelf (dismissed):

src/qfa/api/schemas.py
teeuwski: Let's defer, this touches the other endpoints too.

src/qfa/api/schemas.py
teeuwski: I don't see the reason to include it as this stage.

mariushelf (approved):
LGTM!

PR #70, opened 2026-04-28
Refactor/hexagonal cleanup

mariushelf:
clean up and enforce hexagonal architecture.

* pre-commit hook to verify that imports point in the correct direction (outside-in)
* split services module into services and adapters

Wins:
* testability
* structure & maintainability


teeuwski (approved):
Nice :)

PR #71, opened 2026-04-28
Refactor/structured llm calling

teeuwski:
## Better semantic seperation between internal/ external Pydantic models

Change model names to end with `Api` and `Model` for external and internal models respectively. (e.g. `AnalyzeRequestApi` and `FeedbackItemModel`.

## Structured LLM calling
The [complete endpoint ](https://github.com/rodekruis/qualitative-feedback-analysis/blob/3ccc5773e06666b0ee83400ddb9be926227c0b58/src/qfa/domain/ports.py#L29) now takes a `response_model: type[T_Response]` type as an argument, which can be either `str` or `pydantic.BaseModel`.

## Moved LLM-helper functions to LLMClient
- `_check_injection`
- `_check_token_limit`

Additionally, the retry logic is also connected to the `LLMClient.complete` function, instead of `StandardOrchestrator._call_with_retries`, which has been removed.


src/qfa/adapters/llm_client.py
```diff
@@ -39,14 +46,135 @@ def __init__(
...
+
+        for pattern_name, pattern in _INJECTION_PATTERNS:
+            if pattern.search(user_message):
+                logger.warning(
+                    "Prompt injection detected: pattern=%s",
+                    pattern_name,
+                )
+                msg = f"Prompt injection detected pattern={pattern_name}"
+                raise AnalysisError(msg)
+
```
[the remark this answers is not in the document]
teeuwski: Created a new ticket #75

src/qfa/adapters/llm_client.py
```diff
@@ -86,6 +169,9 @@ async def complete(
                 api_version=self._api_version or None,
                 user=tenant_id,
                 timeout=timeout,
+                response_format=response_model
+                if issubclass(response_model, BaseModel)
```
mariushelf, line 172: this silently drops wrong `response_model` input. If the user supplies anything that is not a `BaseModel`, it won't be used, and the user won't get an error.

I'd suggest to pull the check out of the function call and raise an ValueError if it fails.

src/qfa/adapters/llm_client.py
```diff
@@ -111,8 +200,20 @@ async def complete(
...
+            try:
+                parsed_data: T_Response = cast(
+                    T_Response, response_model.model_validate_json(content)
+                )
+            except ValidationError as exc:
+                raise LLMError(
+                    f"LLM response validation failed for {response_model.__name__}: {exc}"
+                ) from exc
+        else:
+            parsed_data = cast(T_Response, content)
```
mariushelf: This is a no-op at runtime. `cast` only tells the static type checker, hey, `content` is a `T_Response`.
So at runtime nothing happens. 
IMHO I'd still raise when the condition above is not met.

src/qfa/adapters/llm_client.py
mariushelf: And maybe allow `None` for `response_model` so that we can still do unstructured calls.

src/qfa/api/routes.py
```diff
@@ -10,36 +10,36 @@
...
-    AssignCodesResponse,
-    CodeItem,
-    CodeItems,
-    FeedbackItemSummary,
-    HealthResponse,
-    SummarizeAggregateResponse,
-    SummarizeFeedbackMetadata,
-    SummarizeRequest,
-    SummarizeResponse,
+    AggregateSummaryApi,
```
mariushelf: I like the explicit distinction between domain models and api objects.
I don't like that everything is now called XXXApi -- that reads to me as if XXXApi is the Api for XXX. 
What do you think about an `Api` prefix instead of suffix?

The `Model` suffix reads natural -- most of those classes are actually domain models.

src/qfa/domain/models.py
```diff
@@ -42,24 +42,20 @@ class AnalysisRequest(BaseModel):
...
 
 
-class AnalysisResult(BaseModel):
+class AnalysisResultModel(BaseModel):
     """The result of a feedback analysis."""
 
     model_config = ConfigDict(frozen=True)
 
     result: str = Field(description="Analysis output text.")
-    model: str = Field(description="LLM model used for analysis.")
```
mariushelf, line 51: where do those fields live now?

src/qfa/domain/ports.py
```diff
@@ -21,9 +21,11 @@ async def complete(
         self,
         system_message: str,
         user_message: str,
-        timeout: float,
         tenant_id: str,
-    ) -> LLMResponse:
+        response_model: type[T_Response],
+        anonymize: bool = True,
```
mariushelf: i believe I saw a `deactivate_anonymization` or so above. Shall we call both parameters the same, and make them both have the same logic (either both positive, or both negative, but not one "enable=True" and the other "disable=False". Ignore this the two parameters serve different things :)

src/qfa/adapters/llm_client.py
teeuwski: Unstructured calls can be done through passing `str` in the response_model. 
Besides, the `type[T_Response]` is bound to `Basemodel | str`, so the type checker will figure this out for you. Do you recommend adding a defensive check to make sure that we actually receive one of these two types?

src/qfa/domain/ports.py
teeuwski: I removed the anonymize parameter from the `LLMPort`, as currently we're applying it in the `Orchestrator`.

But, the suggestion is still valid. Implementing...

src/qfa/domain/models.py
teeuwski: In `LLMResponse`

src/qfa/adapters/llm_client.py
mariushelf: clear! In that case i think it's fine to skip the check, up to you. But add a docstring that explains how it works :)

mariushelf (approved):
LGTM!

PaulvH63:
Not too much to add really, all good for me 👍

PaulvH63 (approved):

PR #85, opened 2026-05-04
fix(terraform): add -input=false flag to Terraform plan command

teeuwski:
In case the terraform input variables change, the `terraform.yml` workflow would hang waiting for user-input:

image

This tells the workflow that we don't have any input to type in


mariushelf (approved):
LGTM

PR #89, opened 2026-05-07
feat(EspoCRM): add scripts for summarize features

PaulvH63:
## Summary

Adds four EspoCRM scripts that integrate the summarize endpoints of the QFA API into EspoCRM workflows:

- **`set_summarize_aggregate_payload`** — builds the request payload for the `summarize_aggregate` endpoint by looping over related feedback records and constructing the JSON body
- **`set_summarize_per_item_payload`** — same for the `summarize` (per-item) endpoint
- **`translate_summarize_aggregate_reponse_to_ouput`** — parses the API response and formats the aggregate summary (title, quality score as dot-rating + %, summary text) into a readable output string
- **`translate_summarize_per_item_reponse_to_ouput`** — same for per-item summaries; also writes each item's summary back to the `descriptionSummary` field on the `CFeedbackData` record

These scripts are meant to be used as EspoCRM formula scripts in the workflow automation to call and process the QFA summarize features. Both payload scripts include a commented-out test payload for local workflow testing without needing real records.

### DIFF
--- a/scripts/espo_crm/set_summarize_aggregate_payload
+++ b/scripts/espo_crm/set_summarize_aggregate_payload
@@ -1,0 +1,97 @@
+// Script for settings summarize aggregate feature payload in EspoCRM
+
+// Find the related feedback records
+$$backendIDs = record\findRelatedMany('CInsight', id, 'feedbackDatas', 9999,'createdAt', 'desc');
+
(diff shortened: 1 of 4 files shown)


teeuwski (approved):
Nice :)

Issue #1, opened 2026-01-22
[SMALL TICKET] Onboarding focal point Xomnia and kick-off project


LiseMichiels:
### **Who requested this small ticket?**
Lise

### **What needs to be done?**
- [ ] Decide focal point Xomnia
- [ ] Finish onboarding material
- [ ] Onboard focal point
- [ ] Plan kick-off session with Daan/Jacopo

PR #96, opened 2026-05-08
feat(api): api keys in db

### LINKED ISSUE #92: [FEATURE] hold api key hashes in database
Currently, api keys are stored as a json blob in he key vault.

This is secure, but messy and fragile.

It's hard to read, hard to maintain and easy to break.
It does not scale to more than a handful api keys.

Instead: hold api keys in the database.

## Security concerns: 
* do not store plain text keys in the db, but store a salted hash and the salt.
* is it acceptable to manage the keys via the same web app for now? Or do we need to launch a separate container?

## Managing api keys
* via super user endpoint on web app.
* Open questions: db is not publicly accessible. end point needs authentication. This raises a chicken/egg problem. How does the first super user api key get into the DB?
  * maybe a script that ssh's into the container and writes the first key? TBD

teeuwski:
Closes #92


src/qfa/adapters/db.py
```diff
@@ -374,3 +424,144 @@ async def get_all_usage_stats(
...
+                key_id=str(row.key_id),
+                name=str(row.name),
+                hashed_key=SecretStr(str(row.hashed_key)),
+                tenant_id=str(row.tenant_id),
+                is_superuser=bool(row.is_superuser),
+            )
+            if tenant_key.matches_key(provided_key):
+                match = tenant_key
+
+        return match
```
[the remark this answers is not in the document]
teeuwski: Ignoring this for now. Other reviewers are free to disagree

src/qfa/domain/models.py
```diff
@@ -224,10 +226,78 @@ class TenantApiKey(BaseModel):
...
+        without a separate salt column, while still being resistant to
+        brute-force attacks.
+        """
+        return hashlib.scrypt(
+            key.encode("utf-8"),
+            salt=key_id.encode("utf-8"),
+            n=2**14,
+            r=8,
+            p=1,
+        ).hex()
```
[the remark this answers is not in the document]
teeuwski: Ignoring. I think this will suffice for now. Others can, again, disagree.

src/qfa/api/app.py
```diff
@@ -182,12 +189,8 @@ def _resolve_tenant(scope: Scope) -> str:
...
 
-        api_keys = getattr(getattr(app, "state", None), "api_keys", None)
-        if not api_keys:
-            return "anonymous"
-
         try:
-            tenant = validate_api_key(token, api_keys)
+            tenant = await app.state.auth_orchestrator.validate_api_key(token)
             return tenant.name
         except Exception:
```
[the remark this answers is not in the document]
teeuwski: This can be dealt with in a seperate Issue if needed. I would prefer not to implement caching in this PR.

teeuwski:
@copilot I just noticed that I don't have a get enpoint for tenants. Please look at how the other endpoints are configured and create a commit for this new addition. 
Add tests that coincide with how other auth things are tested. Do not add a bunch of helper functions and such, the general structure of the repo is fine.

teeuwski:
@copilot resolve the merge conflicts in this pull request

docs/adr/README.md
```diff
@@ -21,6 +21,7 @@ to them, the options considered, and the reasoning behind the chosen approach.
 | [010](010-shared-container-registry.md) | Shared container registry across environments | Accepted |
 | [011](011-drop-orchestrator-port.md) | Drop swappable-orchestrator requirement and remove OrchestratorPort | Accepted (supersedes [008](obsolete/008-keep-orchestrator-port.md)) |
 | [012](012-postgres-mi-as-admin.md) | App Service managed identity as PostgreSQL Entra admin | Accepted |
+| [013](013-keys-in-db-and-environment-vars.md) | Store API keys in both the database and environment variables | Accepted |
```
mariushelf: this should be in docs/adr/index.md now instead of the README.

src/qfa/adapters/db.py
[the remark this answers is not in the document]
mariushelf: scrypt is designed to run slow, so that cracking it is constrained by cpu. however with at most dozens of keys expected for now, this should not be a problem at the current scale. Agreed to ignore.

src/qfa/adapters/db.py
```diff
@@ -81,6 +92,43 @@ async def _translate_db_errors() -> AsyncIterator[None]:
...
+        sa.DateTime(timezone=True),
+        nullable=False,
+        server_default=sa.func.now(),
+    ),
+)
+
+keys = sa.Table(
+    "keys",
+    metadata,
+    sa.Column("key_id", sa.String(255), primary_key=True),
```
mariushelf, line 141: no expiration date?

src/qfa/adapters/db.py
```diff
@@ -81,6 +92,43 @@ async def _translate_db_errors() -> AsyncIterator[None]:
     sa.Index("idx_llm_calls_timestamp", "timestamp"),
 )
 
+tenants = sa.Table(
+    "tenants",
+    metadata,
+    sa.Column("tenant_id", sa.String(255), primary_key=True),
+    sa.Column("name", sa.String(255), nullable=False),
+    sa.Column("allows_superusers", sa.Boolean, nullable=False, default=False),
```
mariushelf: some documentation for this column would be handy to have here. All other columns are self-explanatory.

src/qfa/adapters/db.py
```diff
@@ -288,7 +336,9 @@ def _select_totals(
     return stmt
 
 
-class SqlAlchemyUsageRepository(UsageRepositoryPort):
+class SqlAlchemyUsageRepository(
+    UsageRepositoryPort, AuthLookupPort, AuthManagementPort
+):
```
mariushelf: The usage repository should have only one concern: record and query usage data.

For authentication you have (rightfully so) created their own ports. There should also be their own repo, e.g. AuthKeyRepository or TenantRepository or something like that. You can choose to have one read-only "repo" (AuthLookupPort) and one write-repo (implementing AuthManagementPort); or combine them into one. At this stage in this repository I'd vote for one combined adapter.

Two adapters have their use, but not here yet (google CQRS)

src/qfa/domain/ports.py
```diff
@@ -157,3 +158,127 @@ def deanonymize(self, text: str, mapping: dict[str, str]) -> str:
...
+
+        Parameters
+        ----------
+        provided_key : str
+            The API key value supplied by the caller.
+
+        Returns
+        -------
+        TenantApiKey
+            The matching tenant API key.
```
mariushelf: ...or None if no matching key is found.

src/qfa/domain/ports.py
```diff
@@ -157,3 +158,127 @@ def deanonymize(self, text: str, mapping: dict[str, str]) -> str:
...
+        """Get all API keys for a tenant, or all keys if tenant_id is None.
+
+        Parameters
+        ----------
+        tenant_id : str | None
+            The tenant to query, or None to get keys for all tenants.
+
+        Returns
+        -------
+        list[dict]
```
mariushelf: Why return a list of dicts instead of the domain model?

src/qfa/domain/ports.py
```diff
@@ -157,3 +158,127 @@ def deanonymize(self, text: str, mapping: dict[str, str]) -> str:
...
+            The API key value to store.
+        key_id : str
+            The unique identifier for the key.
+        key_name : str
+            A human-friendly name for the key.
+        tenant_id : str
+            The tenant this key belongs to.
+        is_superuser : bool
+            Whether this key should have superuser privileges (default False).
+
```
mariushelf, line 253: What does it return?

src/qfa/domain/ports.py
```diff
@@ -157,3 +158,127 @@ def deanonymize(self, text: str, mapping: dict[str, str]) -> str:
...
+        Parameters
+        ----------
+        key_id : str
+            The unique identifier of the API key record to remove.
+        """
+        ...
+
+    async def get_tenants(self) -> list[dict]:
+        """Return metadata for all tenants in the implemented adapter.
```
mariushelf, line 280: same remark as above, why not return the Tenant domain model?

src/qfa/services/auth_orchestrator.py
```diff
@@ -0,0 +1,162 @@
+"""Application service for authentication and management operations."""
+
+from qfa.domain import AuthenticationError, TenantApiKey
+from qfa.domain.ports import AuthLookupPort, AuthManagementPort
+
+
+class AuthOrchestrator:
+    """Coordinate key and tenant lookup and management across configured ports."""
```
mariushelf, line 9: This is a very thin wrapper around the AuthManagementPort. I'd remove the wrapper-methods, and use the AuthManagementPort directly.

The other functions mostly match the AuthLookupPort interface. So you could make the `AuthOrchestrator` itself inherit from `AuthLookupPort`. Then you would wire it once, and then use this composed adapter. Same behaviour as now, but with a nice and explicit interface.

src/qfa/adapters/db.py
mariushelf: Separating the auth ports from the usage repo also allows us to decide whether we want to group the adapters by technology (db vs env), or by feature (usage vs auth).

src/qfa/api/app.py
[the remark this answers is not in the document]
mariushelf: Agreed.

src/qfa/adapters/db.py
```diff
@@ -374,3 +424,164 @@ async def get_all_usage_stats(
...
+                raise TenantDoesNotAllowSuperUsersError(
+                    f"Tenant '{tenant_id}' does not allow superuser keys"
+                )
+
+            try:
+                await session.execute(
+                    keys.insert().values(
+                        key_id=key_id,
+                        name=key_name,
+                        hashed_key=TenantApiKey.hash_key(api_key, key_id),
```
mariushelf: IMHO the salt should be a secret, not a user-supplied value.

src/qfa/api/app.py
```diff
@@ -686,14 +726,36 @@ def create_app(*, llm_factory: LLMFactory | None = None) -> FastAPI:
         The fully configured application instance.
     """
     factory: LLMFactory = llm_factory if llm_factory is not None else build_llm_client
+
+    tags_metadata = [
+        {
```
mariushelf, line 740: nice

src/qfa/api/routes_auth.py
```diff
@@ -0,0 +1,137 @@
...
+    ApiTenantsResponse,
+)
+from qfa.domain.models import TenantApiKey
+from qfa.services.auth_orchestrator import AuthOrchestrator
+
+router = APIRouter()
+
+
+@router.post(
+    "/v1/auth/tenants",
```
mariushelf: now this is a bit picky, but even though I know what the `auth` route is about, I somehow expected to see something like "login" under "auth". Should we rename v1/auth/... to v1/admin/... or so? You can also say tell me to RTFM instead of renaming this :)

src/qfa/api/schemas.py
```diff
@@ -377,6 +377,95 @@ class ApiAssignCodesResponse(BaseModel):
...
+class ApiTenantsResponse(BaseModel):
+    """Response body for ``GET /v1/auth/tenants``."""
+
+    tenants: list[ApiTenant] = Field(description="Tenant metadata records.")
+
+
+class ApiAddKeyRequest(BaseModel):
+    """Request body for ``POST /v1/auth/keys``."""
+
+    api_key: str = Field(
```
mariushelf: this is philosophical, but: do we want to support user-generated keys, or use server-generated keys? Any major service that allows users to access their api *generates* the keys server-side, and returns them to the user once. See github access tokens, for example.

That guarantees that the keys follow a certain standard (e.g., fully auto-generated, secure-random etc)

src/qfa/api/schemas.py
```diff
@@ -377,6 +377,95 @@ class ApiAssignCodesResponse(BaseModel):
...
+        min_length=1,
+        description="Plain API key value to store securely.",
+    )
+    key_id: str = Field(
```
mariushelf: if we auto-generate this instead of accepting it as user input, we don't need to capture and handle collisions.

src/qfa/domain/models.py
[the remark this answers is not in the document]
mariushelf: hmm, I *think* security-wise this does not matter. Best-practice-wise, I'm not sure :)
I'm not an encryption/auth expert. How bad is it if the salt is known client-side, and client chosen?

No pepper is fine.

src/qfa/domain/models.py
```diff
@@ -224,10 +226,78 @@ class TenantApiKey(BaseModel):
...
+
+    @model_validator(mode="before")
+    @classmethod
+    def _normalize_key_inputs(cls, data: Any) -> Any:
```
mariushelf, line 300: can you add a short docstring with the intention of this function?

mariushelf (changes requested):

mariushelf:
Ideally we'd also add/adjust documentation on how authentication works. The ADR already explains WHY we do it the way we do.

src/qfa/api/schemas.py
teeuwski: I like the idea of generating them as well. I haven't given this any thought as of yet.

mariushelf:
Here is and excerpt from Claude's review.

I'd vote for server-side generated keys (my own comment above), and then number 1 and 3 from below. Everything else is probably good-enough for the current state and scope of the project.


## 3. Top-N fix-first

**Ordering rationale.** I order by *blast-radius reduction per change*: how many distinct findings collapse, and whether the fix is structural (prevents whole categories from recurring) or local. The first three are structural and each closes 4–6 findings. They also gate the others — fixing the wire format first (#1) makes the linear-scan fix (#2) trivial and reveals the salt-vs-identity split (#3) cleanly. The PATCH-and-audit items (#4, #5) are independent operational unblocks. The remaining items are cheap point fixes.

| # | Action | Closes | Blast radius |
|---|---|---|---|
| 1 | **Embed `key_id` in the wire format** (e.g. `qfa_<key_id>.<secret>`) so `validate_api_key` is one indexed lookup. | S1, S5, DAT-10, AUTH-04, AUTH-05, S2 | Surfaces the `idx_keys_hashed_key` index as useful (closes S8 as a side effect), enables removing the env-fallback linear-scan path. Reveals: per-`key_id` offline brute-force window — mitigated by #3. |
| 2 | **Scope superusers to a tenant by default**; gate cross-tenant ops on an explicit `is_global_admin` (or `superuser_scope`) flag. Block deletion that would zero the global-admin count. | S6, AUTH-10, AUTH-11 (partial) | Closes the multi-NS cross-tenant escalation. Surfaces: AUTH-09 (no bootstrap path) — addressed by #5. Masks nothing. |
| 3 | **Split `key_id` (public handle) from `salt` (random per-row column)**; composite-PK keys as `(tenant_id, key_id)`. | API-10, DAT-05 (partial), DAT-06, AUTH-02, H1, H2, DAT-14, AUTH-03 | Closes the salt-is-identity-is-PK entanglement. Surfaces: migration work, but the table is brand-new (no production data yet). |
| 4 | **Resolve env-vs-DB precedence explicitly**: fail-fast at startup if any `key_id` appears in both lookups; namespace env keys (`env:<key_id>`); annotate each `get_auth_keys` row with its backend; deduplicate. Document the chain in the composition root. | C1, C2, AR-05, AR-06, API-21, API-25, AUTH-17 | Closes the silent privilege-drift cluster. Masks nothing. Side benefit: removes the silently-merged duplicate rows in `GET /v1/auth/keys`. |
| 5 | **Add `PATCH /v1/auth/tenants/{id}` for `allows_superusers` and `PATCH /v1/auth/keys/{id}` for `is_superuser`/`name`; add a CLI `qfa auth bootstrap` for first-superuser provisioning.** | API-02, G2, AUTH-09, AUTH-11 | Operability: removes the "DELETE-cascade-to-flip-a-flag" footgun. Masks nothing. |
| 6 | **Wrap auth-path DB calls in `_translate_db_errors`; add startup probe; distinguish `IntegrityError` SQLSTATE 23505 vs 23503 in `add_key`.** | G1, AR-15, B1, DAT-13 | Closes "DB outage → 500-storm" and "vanished tenant masquerading as duplicate key". Masks nothing. |
| 7 | **Apply `Field(strict=True)` to all boolean security flags; `model_config = ConfigDict(extra="forbid")` on all `ApiAdd*Request` models; explicit `Path(..., min_length=1, max_length=255)` on path params; NFC-normalise `key_id` and `api_key` before hashing.** | F1, F2, DAT-01, DAT-02, DAT-04, DAT-05 (full closure), F3 (partial) | Closes the type-coercion cluster. Masks nothing. |
| 8 | **Structured audit-log every mutation on `/v1/auth/*` and every failed validation; include `key_id` in `LLMCallRecord` (or document why not).** | S7, AUTH-08, API-22 | Closes the "no observability" cluster. Required to verify that fixes #1–#7 worked in production. |

A ninth, lower-priority but cheap item: **fix the `TenantDoesNotAllowSuperUsersError` → 422/409 mapping** (AR-14, API-15) and **update README** for `/v1/auth/*` (API-27). Both are one-line / one-file changes that materially improve day-1 integrator experience.

src/qfa/api/routes_auth.py
teeuwski: I agree with this. /admin/ also signifies users that they should have admin priviliges to hit certain endpoints. I'll change it, but keep the labels (Default, Usage, User Management, Usage Tracking) the same.

src/qfa/adapters/db.py
teeuwski: I don't particularly like to implement this now. Also introduces work for the managers in the future that I'm not fond of.

src/qfa/adapters/db.py
teeuwski: Resolving in favour of randomly generating api keys [discussion](https://security.stackexchange.com/questions/209936/do-i-need-to-use-salt-with-api-key-hashing)

src/qfa/services/auth_orchestrator.py
teeuwski: I'm unsure how to implement this. Let's discuss later.

teeuwski:
@copilot resolve the merge conflicts in this pull request

src/qfa/services/auth_orchestrator.py
mariushelf: we'll just leave it as it is. Resolving 👍

mariushelf (approved):
LGTM now with your changes and our discussion!

We decided against using a salt, because rainbow table attacks against random keys are ineffective anyway.

Issue #88, opened 2026-05-06
[BUG] Invalid JSON error for summarize endpoint


teeuwski:
### **Describe the bug**
The summarize endpoint (and possibly others) will give the following 502 response:

```json
{
  "error": {
    "code": "analysis_unavailable",
    "message": "LLM returned invalid JSON for summary output",
    "request_id": "req_2JOsvqSd12FoTuZtmjTj4A",
    "fields": null
  }
}
```

### **To Reproduce**
Run the standard input for the summarize endpoint. 

### **Screenshots**

<img width="695" height="569" alt="Image" src="https://github.com/user-attachments/assets/7ce06886-a9d6-4ea2-8d71-33ab922b3c05" />

PaulvH63:
Unable to reproduce the original 502 after investigation.

**Why it no longer occurs 3ccc577:**

_Before:_ The orchestrator manually parsed the LLM's raw text response as JSON. On failure it raised `AnalysisError("LLM returned invalid JSON for summary output")` → 502 `analysis_unavailable`.

_After:_ Structured output (`response_format=SummaryResultModel`) was moved into the LLM client. The model is now forced to return valid JSON directly; parse failures surface as `LLMError` → 502 `llm_error`. The old `AnalysisError` raise no longer exists.

**While investigating I did find two related bugs** in the summarize flow that have been addressed separately:
- #106 LLM was returning wrong record IDs in summarize responses (trusting LLM output instead of pinning IDs from the request). I think that the introduction of Mistral as a replacement of GPT caused this to happen.
- #104 a failing record update call in EspoCRM was blocking the summarize-per-item flow, causing it to not show any results in Espo.

@teeuwski I suggest to close this issue, agree?

teeuwski:
Yes, thanks for the effort! I'll close it now.

PR #103, opened 2026-05-11
fix(espo): remove record update call that blocked summarize per item

PaulvH63:
## Summary

- Removes a `record\update` call inside the Espo CRM formula script that was left over from a previous experimentation and caused the summarize-per-item feature to fail silently on Espo Dev.

## Notes

This branch is named `pvh/fix/json_error_summarize_endpoint` but the actual fix is unrelated to a JSON error in the summarize endpoint. I was unable to reproduce issue #88; this fix addresses a separate problem I did encounter on Espo Dev.

## Test plan
- [ ] Deploy to Espo Dev and verify summarize-per-item runs without errors


PaulvH63:
Replaced by cleaner branch — see new PR.

PR #106, opened 2026-05-11
fix(summarize): pin record IDs from request instead of trusting LLM output

PaulvH63:
## Summary
- Per-item summarize: route handler now zips request records with result summaries so `id` always comes from the caller, not the LLM
- Summarize-aggregate: orchestrator overrides the `ids` field after the LLM call for the same reason

## Why
Some models (e.g. Mistral) return wrong or missing IDs — either echoing sequential numbers (`1`, `2`) or the raw text content of the records (like daan's contact info haha). This caused EspoCRM to receive incorrect IDs, breaking record updates.

### DIFF
--- a/src/qfa/api/routes.py
+++ b/src/qfa/api/routes.py
@@ -149,12 +149,14 @@ async def summarize(
     return ApiSummarizeResponse(
         summaries=[
             ApiFeedbackRecordSummary(
-                id=summary.id,
+                id=record.id,
                 title=summary.title,
                 summary=summary.summary,
                 quality_score=summary.quality_score,
             )
-            for summary in result.feedback_record_summaries
+            for record, summary in zip(
+                body.feedback_records, result.feedback_record_summaries
+            )
         ],
         used_anonymization=body.anonymize,
     )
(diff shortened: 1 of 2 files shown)


src/qfa/api/routes.py
```diff
@@ -149,12 +149,14 @@ async def summarize(
...
         summaries=[
             ApiFeedbackRecordSummary(
-                id=summary.id,
+                id=record.id,
                 title=summary.title,
                 summary=summary.summary,
                 quality_score=summary.quality_score,
             )
-            for summary in result.feedback_record_summaries
+            for record, summary in zip(
```
mariushelf: zip silently drops elements if list lengths do not match. There should at least be a sanity match to check that the list created by the LLM matches the length of the input list.

mariushelf (changes requested):

src/qfa/api/routes.py
mariushelf: question is whether this should happen here, or in the orchestrator. I'd argue the orchestrator -- it owns the business logic, and should check for correctness. Here in the route it's then just format conversion of guaranteed-correct output data

PaulvH63: Updated implementation in https://github.com/rodekruis/qualitative-feedback-analysis/commit/0ade495868489e6291f49995613534a099424c5e

What do you think? @mariushelf

mariushelf: two comments on that commit.

mariushelf (approved):

teeuwski (approved):

PR #108, opened 2026-05-12
Feat/fixes for devcontainer

mariushelf:
* add `dczsh` to open zsh in the container
* firewall fixes so that context7 and huggingface (for memsearch model download) are reachable
* pass CLI args through to claude
* add missing dependencies for memsearch and pyright claude plugins


.devcontainer/.env.example
```diff
@@ -3,3 +3,4 @@
 # This is by design (GH_TOKEN is required for git push etc.).
 # But it means that you need to be careful what you expose here.
 GH_TOKEN=github_pat_...
+CONTEXT7_API_KEY=ctx7sk-...
```
PaulvH63, line 6: Note sure which key this is. 

EDIT: Found definition in README, might be useful to add here that it's optional. Up to you

PaulvH63 (approved):

Issue #91, opened 2026-05-07
expose cost per operation on `usage` endpoints


mariushelf:
Currently, the usage endpoints only group by tenant.

We also want to group by operation / api endpoint that triggered the calls.

The data is already stored in the database. `llm_calls.operation` column.

We might want to add another column `api-call-id` or so, for operations that make multiple calls to the LLM, so that we can aggregate the total cost per api call, not individual LLM calls.

## Output

For a tenant object, add a nested list of calls per operation, with the same metrics as on tenant level.

```json
{
  "tenant_id": "marius-dev",
  "total_calls": 48,
  "failed_calls": 3,
  "total_cost_usd": ...,
  "call_duration": {
    "avg": 2254.5,
    "min": 1776.0,
    "max": 2904.0,
    "p5": 1821.0,
    "p95": 2807.7
  },
  "input_tokens": {
    "avg": 197.75,
    "min": 158.0,
    "max": 211.0,
    "p5": 165.95,
    "p95": 211.0,
    "total": 791
  },
  "output_tokens": {
    "avg": 83.5,
    "min": 80.0,
    "max": 85.0,
    "p5": 80.6,
    "p95": 85.0,
    "total": 334
  },
  "from": null,
  "to": null
},
"operations": [
 {
   "operation": "summarize",
  "total_calls": 4,
  "failed_calls": 0,
  "total_cost_usd": 0.00085,
  "call_duration": {...  },
  "input_tokens": { ...},
  "output_tokens": { ...},
},
 {
   "operation": "analyze",
  "total_calls": 42,
  "failed_calls": 3,
  "total_cost_usd": 42.69,
  "call_duration": {...  },
  "input_tokens": { ...},
  "output_tokens": { ...},
 }
]
}
```

## Implementation
* Reading: add / modify methods on `UsageRepositoryPort` and its implementation(s).
* API: modify `UsageStats` class, or turn it into a base class for TenantUsageStats (which has the tenant_id and operations list), and OperationUsageStats (which has operation_name).

## To be refined
* is the above shape what we want @DGorsseRedCross ? Or do we want it the other way round: Operations top level? With a tenant list? Or not even a tenant list, but only top-level consumption per operation?

## Refinement outcome: we want both: 
* [ ] list of tenants with nested cost per operations, AND
* [ ] list of operations, with nested cost per tenant (superuser access only)

PR #63, opened 2026-04-22
feat: Add summarize-aggregate endpoint

PaulvH63:
Implement summarize-aggregate endpoint.
- Input: list of feedback items
- Output: one summary that catches overall summary of batch of feedback items, sorted on importance.

Closes #37

PR #114, opened 2026-05-19
feat(persistence): add call_id correlation to llm_calls (#91)

mariushelf:
## Summary

Adds a `call_id` UUID column to `llm_calls` so multiple LLM calls made within a single API invocation share one correlation ID, unified end-to-end with the HTTP `X-Request-ID`. Enables future per-invocation cost aggregation on `/v1/usage` (see #91). API/endpoint output shape is intentionally still deferred — being refined in #91.

Request-id and call context are now created by depending on `call_scope_for` in every route that requires a call scope. The Orchestrator does not to wrap everything into an `async with call_scope` anymore. See `docs/architecture/04-crosscutting.md`.

## What changed

- **Domain** — `CallContext` and `LLMCallRecord` gain a required `call_id: UUID` field.
- **Scope handling** — `call_scope(tenant_id, operation, request_id)` lives in `qfa.services.call_context`, backed by a single `current_call_context: ContextVar[CallContext | None]`. The orchestrator no longer enters scope; a FastAPI dependency `call_scope_for(operation)` at the driving-adapter layer does, reading `request.state.request_id` set by `RequestIdMiddleware`.
- **Unification with X-Request-ID** — `RequestIdMiddleware` generates one `uuid4()` per request and writes it to `request.state.request_id` and the `X-Request-ID` response header. `call_scope_for` passes that same UUID into `call_scope(..., request_id=...)`, which becomes `ctx.call_id`. Header value = body `request_id` = every `llm_calls.call_id` row from that request.
- **Adapter** — `TrackingLLMAdapter` copies `ctx.call_id` into every persisted record on both success and error paths; refactored to a single `inner.complete` call with a discriminated-union outcome (`LLMResponse | Exception`) feeding one `_build_record` helper. Fail-open: missing `call_scope` logs at ERROR and bypasses persistence, never breaks the user-facing call.
- **Persistence** — `llm_calls` table gains a non-null `call_id` UUID column (native UUID on Postgres) plus composite index `idx_llm_calls_tenant_operation_call_id` for the future aggregation query.
- **Migration** — `20260519_01_add_call_id_to_llm_calls`: adds the column nullable, backfills each existing row with a distinct `gen_random_uuid()`, then enforces `NOT NULL`. Old rows look like one-LLM-call invocations.
- **Docs** — `docs/architecture/04-crosscutting.md` rewritten to describe the unified scope mechanism with a 3-box structural mermaid diagram (route writes → `current_call_context` ← TrackingLLMAdapter reads). `docs/architecture/03-components.md` and `docs/architecture/05-data-model.md` updated accordingly.
- **Tests** — new unit tests for context generation, fan-out correlation (`asyncio.gather`), DB round-trip (SQLite + Postgres), and migration backfill. New e2e test `test_fanned_out_llm_calls_share_one_call_id_equal_to_request_id` verifies the full unification chain end-to-end through `/v1/summarize-aggregate`.

## Test plan

- [x] `make test` → unit + service tests green
- [x] `make lint` → ruff + ty + import-linter clean
- [x] `make docs` → Sphinx clean
- [x] `make db-up && make test-integration` → integration + e2e green (run on host, includes new e2e unification test)

## Review focus

- **`call_scope` is entered at the driving-adapter layer, not the orchestrator.** Each route declares its operation via `Depends(call_scope_for(Operation.X))`. The orchestrator is pure use-case logic; the tracking adapter reads the ambient context. Worth a look in `src/qfa/api/routes.py` to confirm the inline `Depends(...)` pattern reads cleanly.
- **Fail-open on missing scope.** `TrackingLLMAdapter` logs at ERROR and routes through to the inner LLM rather than raising. Observability never breaks the use case; missing scope is alertable but does not fail user-facing requests. See `tests/adapters/test_tracking_llm.py::test_bypasses_persistence_and_logs_when_call_scope_unset`.

## Follow-up

- Route-naming inconsistency observed while debugging the e2e test: `/v1/assign_codes` uses underscores, `/v1/summarize-aggregate` uses a hyphen. Both were introduced by the same author one day apart (#56 and #63); appears to be accidental drift. Worth a dedicated issue to standardize — out of scope here. See #115.


src/qfa/adapters/db.py
```diff
@@ -58,6 +58,7 @@ async def _translate_db_errors() -> AsyncIterator[None]:
     ),
     sa.Column("tenant_id", sa.String(255), nullable=False),
     sa.Column("operation", sa.String(64), nullable=False),
+    sa.Column("call_id", sa.Uuid(), nullable=False),
```
mariushelf: add inline-documentation here

src/qfa/services/call_context.py
```diff
@@ -32,13 +34,21 @@ async def call_scope(
...
     Yields
     ------
     CallContext
         The context that was set.
     """
-    ctx = CallContext(tenant_id=tenant_id, operation=operation)
+    ctx = CallContext(
+        tenant_id=tenant_id,
+        operation=operation,
+        call_id=call_id if call_id is not None else uuid4(),
```
mariushelf: extract into its own line before `ctx`. Do not mix object instantiation with logic.

mariushelf:
All test functions that you created or modified should have *at least* a one-line docstring describing what is tested and why.

src/qfa/adapters/db.py
mariushelf: Added an inline comment on the column explaining what `call_id` correlates and how `/v1/usage` will use it ([d595472](https://github.com/rodekruis/qualitative-feedback-analysis/commit/d595472)).

~Claude

src/qfa/services/call_context.py
mariushelf: Extracted the ternary into a `resolved_call_id` local before the `CallContext(...)` call ([d595472](https://github.com/rodekruis/qualitative-feedback-analysis/commit/d595472)). Construction now reads as pure data assembly with no logic inline.

~Claude

mariushelf:
Added one-line docstrings (what + why) to every test function I created or modified for this PR, across `tests/services/test_call_context.py`, `tests/domain/test_models.py`, `tests/adapters/test_db.py`, `tests/adapters/test_tracking_llm.py`, `tests/integration/test_db_postgres.py`, and `tests/integration/test_migrations.py` ([d595472](https://github.com/rodekruis/qualitative-feedback-analysis/commit/d595472)). Saved the convention to memory so I apply it by default next time.

~Claude

src/qfa/adapters/db.py
mariushelf: use `doc` or `comment` argument on constructor instead

mariushelf: Switched to `sa.Column(..., comment=...)` on both the schema definition and the Alembic `add_column` call ([51e6f41](https://github.com/rodekruis/qualitative-feedback-analysis/commit/51e6f41)). The documentation now propagates to the live DB as a `COMMENT ON COLUMN`, visible via `\d+ llm_calls` in psql and queryable from `pg_description`.

~Claude

src/qfa/adapters/tracking_llm.py
```diff
@@ -57,17 +56,29 @@ async def complete(
...
+        invoked outside an HTTP request (e.g. a CLI or test that forgot
+        to set up scopes); HTTP paths set the scope via
+        ``call_scope_for`` at the route layer.
         """
         ctx = current_call_context.get()
         if ctx is None:
-            raise MissingCallScopeError(
-                "TrackingLLMAdapter.complete called outside an active call_scope; "
-                "the orchestrator must enter call_scope(...) at each public-method entry."
+            logger.error(
```
mariushelf: I suggest instead:

```
try:
  self._inner.complete(...)
  status=CallStatus.OK
except Exception:
  status=CallStatus.ERROR
if ctx is None:
   log error
else:
  self._record_safely()
```
Less duplication.

mariushelf: Done in [ec099bb](https://github.com/rodekruis/qualitative-feedback-analysis/commit/ec099bb). Captured the inner-LLM outcome into a single `LLMResponse | Exception` local, then branch once: log on missing scope, build + record otherwise. Extracted `_build_record` so the shared 5 fields (`tenant_id`, `operation`, `call_id`, `timestamp`, `call_duration_ms`) live in one place; `isinstance(outcome, Exception)` narrows the type without runtime asserts.

~Claude

src/qfa/services/orchestrator.py
```diff
@@ -22,12 +22,10 @@
     CodingAssignmentRequestModel,
```
mariushelf, line 22: Note to reviewer: the only intended change here is to remove the `async with call_scope` from all functions. Since this changes indentation, the diff looks scary. But it's really just 4 lines removed.

teeuwski (approved):
I like this change a lot, makes it harder to make mistakes and removes indentation + repetition

PR #120, opened 2026-05-20
feat(anonymize): add multi-language anonymization models

### LINKED ISSUE #118: feat(anonymization): implement translation for 5 common languages
Implement the anonymization flow for 5 common languages:
- English
- Spanish
- Ukrainian
- Russian
- French

The flow consists of:
1. Detecting the language
1. Using a SpaCy entity recognition model specific to the detected language

Use a fallback multi-language model in case the detected language does not have a downloaded model associated with it.

teeuwski:
Closes #118


scripts/test_pii_removal.ipynb
```diff
@@ -34,20 +32,19 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "anonymized_string, mapping = orch.anonymize(\n",
-    "    \"Amsterdam is my city. Did you hear me, Amsterdam is my city! My name is John\"\n",
-    ")\n",
-    "anonymized_string, mapping"
```
mariushelf, line 40: I think this should be a test and not a notebook.

Edit: aah, it already is in the tests file. Do you think it's worth keeping this NB then? Or can it be removed? I'd remove it, but I don't have a strong opinion on it :)

pyproject.toml
```diff
@@ -35,8 +35,23 @@ dependencies = [
...
+    "fr_core_news_sm",
+    "uk_core_news_sm",
+    "ru_core_news_sm",
+    "es_core_news_sm",
+    "xx_ent_wiki_sm",
+    "langdetect>=1.0.9",
 ]
 
+[tool.uv.sources]
+en-core-web-sm = { url = "https://github.com/explosion/spacy-models/releases/download/en_core_web_sm-3.8.0/en_core_web_sm-3.8.0-py3-none-any.whl" }
```
mariushelf, line 48: do we add these sources because the models are not released on PyPI? I think it's worth a short comment to explain.

tests/adapters/test_presidio_anonymizer.py
```diff
@@ -40,3 +44,44 @@ def test_date_entities_are_preserved_verbatim(
...
     assert len(mapping) == 0
+
+
+def test_anonymize_handles_undetectable_language_gracefully(
+    anonymizer: PresidioAnonymizer,
+) -> None:
+    # This string is just random characters unlikely to be detected as any language
+    input_text = "asdkjhasdkjh asdkljhasd asdkljhasd"
+    anonymized_text, mapping = anonymizer.anonymize(input_text)
+    assert anonymized_text == input_text  # No anonymization should occur
```
mariushelf, line 55: Ok, this tests the case where "xx" maps to "no language". So input==output is correct here. Just thinking out loud :)

tests/adapters/test_presidio_anonymizer.py
```diff
@@ -16,6 +16,10 @@ def anonymizer() -> PresidioAnonymizer:
         ("Hi my name is Dick Schoof", "<PERSON", "Dick Schoof"),
         ("My number is 212-555-5555", "<PHONE_NUMBER", "212-555-5555"),
         ("I live in The Netherlands", "<LOCATION", "The Netherlands"),
+        ("Je m'appelle Marie Dupont", "<PERSON", "Marie Dupont"),  # French
+        ("Me llamo Carlos García", "<PERSON", "Carlos García"),  # Spanish
+        ("Мене звати Олена Петренко", "<PERSON", "Олена Петренко"),  # Ukrainian
+        ("Меня зовут Иван Иванов", "<PERSON", "Иван Иванов"),  # Russian
```
mariushelf, line 22: Here we could add a case where "xx" maps to an actual language. E.g. "Ciao, sono Marco. Abito a Roma. " should replace Marco and Roma with the wiki model. Right?

src/qfa/adapters/presidio_anonymizer.py
```diff
@@ -5,11 +5,40 @@
...
+    """
+    try:
+        language_shortcode = detect(text)
+    except LangDetectException:
+        return "xx"
+
+    if language_shortcode not in [
+        ent["lang_code"] for ent in LANGUAGES_AND_ANONYMIZATION_MODEL_PAIRINGS
+    ]:
+        return "xx"
```
mariushelf, line 39: I wonder if we should leave the translation from "non-listed language" to "xx" to the caller. It would allow us to e.g. log "warning: unsupported language: IT" or so, and distinguish between "undetected language and unsupported language". Admittedly only relevant for logging at the moment.

mariushelf (dismissed):
LGTM.

mariushelf:
All in all, LGTM.

Just some remarks on notebook philosophy and observability:
* notebooks: when do we keep them, when do we remove them? If their only purpose was to experiment with new tech, but by now we have test cases and a working implementation, I'd vote for "delete". If they are useful to showcase how to use certain features of tech, and it's referenced from the docs: keep.
* observability: do we want to fold all unknown and undetected languages into one, or do we want to be able to log the used language? I'd vote for the latter, but for now, either is fine.

tests/adapters/test_presidio_anonymizer.py
teeuwski:
```bash
FAILED tests/adapters/test_presidio_anonymizer.py::test_anonymize_replaces_entity_and_roundtrip_restores_original[Ciao, sono Marco. Abito a Roma.-<PERSON-Marco] - AssertionError: assert '<PERSON' in '<LOCATION_3>, sono <LOCATION_2>. <LOCATION_1> a <LOCATION_0>.'
```

Tried this here 😆, it's really a "last resort" model. This is getting closer to a Data Science model evaluation.

mariushelf (approved):
LGTM!

PR #132, opened 2026-05-21
refactor(tests): inherit ports explicitly in test doubles

mariushelf:
## Summary

- Tighten the project's port-inheritance rule: explicit inheritance is now the **default for every class that implements a port**, including test doubles. Only genuinely ad-hoc cases (e.g. `MagicMock(spec=Port)`) skip it.
- Updates `AGENTS.md`, `docs/adr/002-protocol-based-ports.md`, and `docs/development/index.md` to state the new default consistently.
- Adds the missing `(LLMPort)` / `(UsageRepositoryPort)` / `(AnonymizationPort)` / `(AuthManagementPort)` bases on 8 test fakes that previously relied on structural conformance.
- `RecordingAuthOrchestrator` in `tests/api/test_auth_routes.py` now subclasses the concrete `AuthOrchestrator` service so `ty` enforces LSP-compatible overrides — signature drift in the service fails the test build. This also surfaced (and fixed) a latent drift where the spy's `validate_api_key` mirrored the lookup *port* (returned `None` on miss) instead of the service contract (raises `AuthenticationError`).

## Test plan

- [x] `uv run pytest tests -q` — 287 passed
- [x] `uv run ruff format --check` / `uv run ruff check` — clean
- [x] `uv run ty check` — clean
- [x] `uv run lint-imports` — 3/3 contracts kept
- [x] `uv run pytest --doctest-modules src` — no doctests collected (parity with main)

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -34,14 +34,18 @@ Hexagonal architecture. Key concepts:
 - Driven adapters (LLM provider, anonymisation) sit behind ports
   declared in `qfa.domain.ports` — for example `LLMPort` and
   `AnonymizationPort` — so implementations can be swapped.
-- **Adapter classes must explicitly inherit from their port** (e.g.
-  `class LiteLLMClient(LLMPort):`, `class PresidioAnonymizer(AnonymizationPort):`).
-  Although Python `Protocol`s support structural typing without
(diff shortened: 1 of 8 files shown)


teeuwski (approved):
Easy change to approve :)

PR #134, opened 2026-05-27
feat(analysis): hierarchical map-reduce analysis for large corpora (#124)

### LINKED ISSUE #124: Free-text analysis: support large corpora beyond single-call token limit
Follow-up to #117. V1 of the free-text `/v1/analyze` endpoint enforces a hard token cap (`FeedbackTooLargeError` / 422) when the batch + system prompt exceed the configured limit. This is sufficient for short-horizon questions ("themes this week in region X") but does not scale to:

- Longitudinal questions ("how have rumours evolved over six months?")
- Cross-region / cross-channel aggregation over thousands of records
- Bulk re-analysis after a coding framework change

## Goal

Add a way to run free-text analysis over corpora that exceed the single-call token budget, without compromising the V1 hard cap as the default.

## Reserved API surface (already in place)

V1 reserves `mode: "single_pass"` on the request. This ticket adds at least one additional value, e.g. `mode: "hierarchical"`.

## Approaches to evaluate (no decision yet)

1. **Map-reduce / hierarchical summaries-of-summaries** — chunk records into token-fitting groups, apply the analyst's prompt to each chunk to produce a partial answer, then synthesise the partials. Requires a distinct leaf prompt and synthesis prompt; faithfulness degrades at each level. Quality scoring for hierarchical mode is non-trivial (the top-level judge cannot see all input verbatim).

2. **Embedding-based retrieval (RAG-lite)** — embed records, embed the analyst's prompt, retrieve top-K, run analysis on those. Faithful per call but inappropriate for trend / aggregate questions where the analyst expects the *whole* dataset to be considered. Adds infrastructure (embedding model adapter, vector storage).

3. **Two-stage hybrid** — retrieval narrows the corpus, hierarchical synthesis produces the answer.

The chosen approach must be a complete feature, including the embedding ingestion pipeline if approach 2 or 3 is selected.

## Out of scope of this ticket

- Quality / uncertainty calibration for the new modes must be designed alongside the chosen approach, but advanced calibration research is out of scope.

## Acceptance criteria

- Design proposal documenting trade-offs of the three approaches and a recommendation.
- A choice approved before any implementation begins.
- Implementation plus a meaningful test set demonstrating the chosen approach on corpora at least 5× the V1 token cap.
- API change is non-breaking (V1 clients sending `mode: "single_pass"` or no `mode` field continue to work unchanged).
- Quality / uncertainty signalling extended to the new mode.

## References

- Parent: #117
- Related: #75 (prompt injection defence) — both modes need to share the same guardrails contract.
- Architecture: hexagonal layering must be preserved.

mariushelf:
> ## ⚠️ DO NOT MERGE THIS PR ON ITS OWN
>
> **#134 must not reach `main` alone.** Its embedding adapter is knowingly
> broken in this diff (two bugs, documented below) and is only fixed by
> **#135**, which is stacked directly on top of this branch.
>
> **Merge procedure:** merge #134 to `main`, then **immediately** merge #135
> (GitHub auto-retargets #135 to `main` once this lands). Better still, merge
> via the stack tip — `feat/corpus-runner` (#135's head) already contains all
> of #134 — in a single integration, and close this PR as "landed via #135",
> so `main` never sees #134 standalone.
>
> #134 and #135 are kept as **separate PRs only to make review tractable**;
> they ship as **one unit**.
>
> ### The two embedding bugs accepted in this diff (fixed in #135)
> Both are in `src/qfa/adapters/embedding.py`:
> 1. **No tokenizer padding** — the tokenizer is never configured for padding,
>    so `np.array([e.ids …])` raises on any batch of differing-length texts.
> 2. **Wrong pooling** — `embed` mean-pools `outputs[0]` as if it were token
>    embeddings, but the shipped BGE-M3 ONNX head already emits a pooled
>    `dense_vecs` output. #135 consumes the pooled head and L2-normalises it.

## Summary

Implements `mode: "hierarchical"` on `POST /v1/analyze` (#124). For corpora
> 5x the single-call token cap the endpoint now runs an embed → cluster →
map → reduce pipeline rather than failing with 413.

**Base:** this PR targets `main`. It is the bottom of a three-PR stack
(#134 ← #135 ← #144); see the merge warning above.

Closes #124.

## What was added

**Domain**
- `EmbeddingPort` (synchronous `Protocol`) in `qfa.domain.ports`.
- `Chunk` model (`qfa.domain.chunk_models`) — token-budget-sized record group.
- `CodingTrendTable` / `CodingTrendCell` (`qfa.domain.clustering_models`).
- `AnalysisRequestModel.mode` widened to `Literal["single_pass", "hierarchical"]`.
- `AnalysisResultModel` extended with optional `confidence` and `coding_trends`.

**Adapters**
- `BgeM3OnnxEmbedder` (`qfa.adapters.embedding`) — BGE-M3 ONNX-int8, explicit
  inheritance from `EmbeddingPort`, security flags asserted at construction.
  (See the two known bugs called out in the warning above — fixed in #135.)

**Services**
- `cluster_records` (`qfa.services.clustering`) — HDBSCAN + token-budget
  chunking, full-coverage invariant, noise label → uncategorised chunk.
- `build_coding_trend_table` / `render_coding_trend_table`
  (`qfa.services.coding_trends`) — deterministic code-by-period aggregation.
- `build_map_system_message` / `build_reduce_system_message` /
  `build_reduce_user_message` (`qfa.services.hierarchical_prompts`) — guardrails
  in both map and reduce prompts.
- `Orchestrator.analyze_hierarchical` — anonymise → embed → cluster → map
  (with leaf judge) → tree-reduce; coverage-weighted confidence.

**API**
- `ApiCodingTrendCell`, `ApiCodingTrends` schemas; `ApiAnalyzeResponse`
  extended with `confidence` and `coding_trends`.
- Route dispatches `analyze_hierarchical` for `mode="hierarchical"`.
- `build_embedder()` in `app.py` — constructs adapter when `EMBEDDING_MODEL_PATH`
  set, returns `None` otherwise (graceful degradation).

**Tests**
- `tests/services/test_orchestrator_hierarchical.py` — async unit tests:
  full coverage + confidence, guardrails at map and reduce, anonymise-before-embed,
  recursion triggers 1 and 2.
- `fixtures/large_corpus.yaml` + corpus-fixture tests — multilingual synthetic
  records (> 5x cap), covering HDBSCAN, coding-trend counts, and the coverage
  invariant. (Generated by `scripts/generate_corpus.py`; marked
  `linguist-generated` so it collapses in diffs.)

**Docs / ADR**
- ADR-014: EmbeddingPort and self-hosted BGE-M3 ONNX embedding model.
- Updated `docs/architecture/03-components.md`, `docs/rest-api/index.md`,
  `docs/operations/settings-reference.md`.

---


src/qfa/api/routes.py
```diff
@@ -109,7 +115,8 @@ async def analyze(
     -------
     AnalyzeResponse
         The analysis result with quality score, uncertainty explanation,
-        feedback record count, and request ID.
+        feedback record count, and request ID.  ``confidence`` and
+        ``coding_trends`` are populated only for ``hierarchical`` mode.
```
mariushelf: coding_trends could and should be easily populated for single-pass, too, if codes are part of the data, no?

src/qfa/domain/clustering_models.py
```diff
@@ -0,0 +1,42 @@
...
+
+    code: str = Field(description="Coding label (per the coding framework).")
+    period: str = Field(description="Time period bucket, e.g. ``2024-01``.")
+    count: int = Field(
+        ge=0, description="Number of records coded with this code in this period."
+    )
+
+
+class CodingTrendTable(BaseModel):
+    """Deterministic, non-LLM count of codes over time periods.
```
mariushelf, line 44: I have a feeling a polars or numpy table would be more suitable. Both are standard enough to be used in the domain.

src/qfa/services/coding_trends.py
```diff
@@ -0,0 +1,110 @@
...
+"""
+
+from collections import Counter
+from collections.abc import Sequence
+
+from qfa.domain.clustering_models import CodingTrendCell, CodingTrendTable
+from qfa.domain.models import FeedbackRecordModel
+
+
+def _period_of(raw_date: object) -> str | None:
```
mariushelf: raw_date should be annotated as str, no?

src/qfa/services/coding_trends.py
```diff
@@ -0,0 +1,110 @@
...
+    """Return the ``YYYY-MM`` bucket for an ISO-8601-ish date string.
```
mariushelf: we should make the periods configurable. Sometimes only one month of data will be submitted. Then the "buckets" need to be smaller, e.g., weeks or days

src/qfa/services/coding_trends.py
```diff
@@ -0,0 +1,110 @@
...
+
+
+def _codes_in_record(
+    record: FeedbackRecordModel, code_fields: Sequence[str]
+) -> list[str]:
+    """Extract coding labels from a record's metadata.
+
+    Each configured code field may hold a comma-separated string of
+    labels (matching the corpus convention). Empty/missing fields
+    contribute nothing.
```
mariushelf, line 93: document the parameters

src/qfa/services/coding_trends.py
```diff
@@ -0,0 +1,110 @@
...
+
+    Parameters
+    ----------
+    records : tuple[FeedbackRecordModel, ...]
+        The full input record set.
+    date_field : str
+        Metadata key holding the record's date (parsed to a ``YYYY-MM``
+        period).
+    code_fields : Sequence[str]
+        Metadata keys holding coding labels (comma-separated strings).
```
mariushelf, line 140: period length needs to be configurable as stated above. Currently it's always one month, but it might as well be per week or per day even.

src/qfa/settings.py
```diff
@@ -76,6 +91,24 @@ class OrchestratorSettings(BaseSettings):
     retry_jitter_factor: float = 0.5
     retry_cap_seconds: float = 10.0
     chars_per_token: int = 4
+    # Hierarchical-analysis (mode="hierarchical") tuning:
+    min_cluster_size: int = Field(
```
mariushelf, line 117: The orchestrator will be split eventually. So instead of putting this into orchestrator settings, create a new class AnalyzeSettings For anything related to the analyze endpoint.

src/qfa/api/app.py
```diff
@@ -661,12 +696,15 @@ async def lifespan(app: FastAPI) -> AsyncGenerator[None, None]:
         )
         logger.info("Usage tracking enabled (per-attempt, per-operation)")
 
+        embedder = build_embedder(settings.embedding)
```
mariushelf, line 703: add logging before and after building the embedder (sth like "Building the embedding model..." "Done")

src/qfa/api/routes.py
mariushelf: You're right — this was an artificial restriction. The coding-trend table is deterministic post-processing of input metadata; it has no dependency on chunking or on the LLM call. Fixed in 7236cf0: `Orchestrator.analyze` now also calls `build_coding_trend_table` and returns it in `AnalysisResultModel.coding_trends`. The route docstring and the REST API doc are updated to say "populated for both modes whenever metadata permits"; pinned by a new test `test_single_pass_populates_coding_trends_from_metadata` in `tests/services/test_orchestrator.py` so a future refactor can't silently revert it.

~Claude

src/qfa/domain/clustering_models.py
mariushelf: Considered and ultimately deferred — wanted to flag the reasoning so it's on the record rather than silently dropped.

**Why a dataframe would be ergonomic:** agreed. A `(code, period, count)` triple-store is exactly what polars/numpy is built for. Pivots ("table view"), groupbys ("collapse weeks into months"), and serialisation are all one-liners instead of nested-loop helpers. If we add downstream consumers that want to re-aggregate the table, the current `tuple[CodingTrendCell, ...]` representation forces every consumer to re-implement the same Counter logic.

**Why not in this PR:**

1. **Domain layering.** `qfa.domain` currently only depends on `pydantic`. ADR-001 + the import-linter contract pin domain as infrastructure-light. Adding polars (or even numpy) to `qfa.domain` is a meaningful architectural change — even if both are "standard enough", they have install footprint, version-coupling, and wheel-build implications that go beyond pydantic. Worth its own ADR amendment rather than a quiet import.

2. **Wire shape.** The table has to serialise to JSON for the API response anyway. The current Pydantic shape is what flows over HTTP — switching the internal representation doesn't change what clients see, so it doesn't unblock anything that's blocked today.

3. **Best return on review effort.** Comments 1, 3/4/6, 5, 7, 8 are all small, fixable, and unambiguously the right thing to do. The polars/numpy question is genuinely a design discussion that deserves its own thread — not a hurried side-effect of a review-fixup PR.

**Proposed split:** if/when we add consumers that need re-aggregation (e.g. a `/v1/analyze` response that lets the client request a different `period` post-hoc, or a coding-trend visualisation that wants polars pivot semantics), we open a follow-up issue + ADR for "domain dataframe types" and migrate. Until then, the Pydantic wire-shape stays.

Alternative path if you'd like dataframe ergonomics without amending the domain contract: add a `CodingTrendTable.to_polars()` adapter in `qfa.services` (where dataframe deps are allowed) and have consumers call that. Happy to do that here as a separate commit if you'd prefer it now.

~Claude

src/qfa/services/coding_trends.py
mariushelf: Fair point — the previous signature was loose. `raw_date` started life as the raw value pulled out of `record.metadata.get(date_field)`, which is `dict[str, str | int | float | bool]`, so in principle it could be any of those types. The `isinstance(raw_date, str)` guard already handled that, but the `object` annotation made it look like we were expecting genuinely heterogeneous input rather than just "we know it should be a string but we're tolerant".

After the rewrite for the configurable period (7236cf0), the function signature is now `_period_of(raw_date: object, period: TrendPeriod) -> str | None` — I kept `object` deliberately so the function stays the type-defensive boundary that early-returns `None` on anything non-string. The docstring now spells that out explicitly:

> Anything that isn't a string returns ``None``; strings are parsed leniently from their leading characters so ``"2024-01-05T10:00:00Z"`` works the same as ``"2024-01-05"``.

If you'd prefer a stricter `str` annotation with the type narrowing pushed up to the caller (`build_coding_trend_table` would then need to do the `isinstance` check inline before calling `_period_of`), happy to flip it — but the current setup keeps the "best-effort, never raise" contract in one place rather than spreading the guard across two functions.

~Claude

src/qfa/services/coding_trends.py
mariushelf: Done in 7236cf0. Period is now configurable as `day` / `week` / `month`, default `week`, exposed both as a server-side default (`ANALYZE_DEFAULT_CODING_TREND_PERIOD`) and per-request via the `period` field on the request body (`ApiAnalyzeRequest` + `AnalysisRequestModel`). One-month corpora can now pick `day` or `week` to surface trend signal.

Implementation notes:
- `_period_of` dispatches on the period; reuses the same string-parsing approach (no `datetime.fromisoformat` dependency).
- `week` uses ISO week labels (`YYYY-Www`) anchored on `iso_year`, not the calendar year. Subtle but important: `date(2024, 12, 30).isocalendar()` returns `(2025, 1, ...)` — using the calendar year would silently merge late-December 2024 records into "2024-W01" alongside truly-January-2024 records.
- The `AnalysisResultModel.coding_trends.cells[].period` field now carries a shape that depends on the chosen granularity (`YYYY-MM-DD` / `YYYY-Www` / `YYYY-MM`); the API schema docstring spells that out.
- Test coverage: new `test_counts_codes_per_week_period_default` and `test_counts_codes_per_day_period` in `tests/services/test_coding_trends.py`; existing month-bucket assertions explicitly pin `period="month"` so they keep working as the regression test for that branch.
- Backward compatibility: the existing `fixtures/large_corpus.yaml` acceptance test (`test_coding_trend_table_counts_match_corpus`) is pinned to `period="month"` because the fixture's date stamps aren't ISO-week-aligned.

~Claude

src/qfa/services/coding_trends.py
mariushelf: Done in 7236cf0. Added a numpy-style Parameters + Returns block to `_codes_in_record` documenting both the per-field comma-splitting behaviour and the missing-key tolerance (silent skip rather than raise) — both were implicit before. The function is private (`_`-prefixed) but it was the only undocumented function in an otherwise fully-documented module, and the silent-skip behaviour is load-bearing for the "best-effort, never raise" contract — worth pinning in prose.

~Claude

src/qfa/services/coding_trends.py
mariushelf: Same fix as the comment above — period is now a parameter on `build_coding_trend_table` (defaulting to `"week"`), threaded through from the request body's `period` field with `ANALYZE_DEFAULT_CODING_TREND_PERIOD` as the server-side fallback. The function's docstring lists which period to choose for which corpus shape (day for short-window deep-dives, week for the typical 1-3 month operational corpus, month for multi-year corpora).

~Claude

src/qfa/settings.py
mariushelf: Done in 7236cf0. New `AnalyzeSettings` class (env prefix `ANALYZE_`) now carries the four analyze-endpoint-specific knobs (`min_cluster_size`, `clustering_metric`, `coding_trend_date_field`, `coding_trend_code_fields`) plus the new `default_coding_trend_period`. `OrchestratorSettings` is reduced to truly cross-cutting wiring (retry policy, metadata allow-list, `chars_per_token`) — those legitimately apply to every endpoint, not just analyze.

Wiring:
- `Orchestrator.__init__` takes a new optional `analyze_settings: AnalyzeSettings | None = None` parameter (defaults to `AnalyzeSettings()` if omitted, so tests that don't care about analyze tuning don't have to thread it through).
- `AppSettings.analyze` is the composition-root field; the FastAPI lifespan passes it into `Orchestrator(...)`.
- Both `analyze_hierarchical` and the new `analyze` trend-table code path read from `self._analyze_settings.*`.

Settings reference (`docs/operations/settings-reference.md`) now has a dedicated **Analyze** section documenting the new env vars. Naming the group after the endpoint (rather than the mode) means a future hierarchical-only knob — or a future single_pass-only knob — both land here without another rename.

ENV-VAR MIGRATION (heads-up for any existing deployment):
- `ORCHESTRATOR_MIN_CLUSTER_SIZE` → `ANALYZE_MIN_CLUSTER_SIZE`
- `ORCHESTRATOR_CLUSTERING_METRIC` → `ANALYZE_CLUSTERING_METRIC`
- `ORCHESTRATOR_CODING_TREND_DATE_FIELD` → `ANALYZE_CODING_TREND_DATE_FIELD`
- `ORCHESTRATOR_CODING_TREND_CODE_FIELDS` → `ANALYZE_CODING_TREND_CODE_FIELDS`

Pre-1.0, so I didn't add aliasing shims — but happy to add `Field(alias=...)` fallbacks if you'd prefer a graceful transition window. ADR-015 was also updated to reference the new env-var names.

~Claude

src/qfa/api/app.py
mariushelf: Done in 7236cf0. The lifespan now emits the two log lines around `build_embedder`:

```
Loading embedding model from {path} ...
Embedding model ready (hierarchical analysis available)
```

The "loading" line is guarded by `if settings.embedding.model_path:` so the empty-path branch stays quiet (it already has its own `EMBEDDING_MODEL_PATH not set — hierarchical mode requires it at runtime` info log inside `build_embedder`). The "ready" line only fires when the build actually returned an embedder. Same INFO level and same pattern as the existing `"Usage tracking enabled (per-attempt, per-operation)"` line directly above, so it slots into the same log stream a deployment is already tailing.

~Claude

teeuwski (approved):
LGTM, quite a complex thing to merge though- but I can understand the need for it. 

I wonder about an `auto` option that selects which of the two modes (`hierarchical` or `single_pass`) to not make things break for users

PR #137, opened 2026-05-29
feat(espo): unify workflows

### LINKED ISSUE #116: Unify EspoCRM workflows into a single Flowcharts
Currently, we have several EspoCRM scripts.

When updating them, the EspoCRM admin needs to update all those scripts.

Instead, we want a single workflow file which can call and handle all QFA api endpoints.

Probably related: #101 , #110 .

----

Comments from @DGorsseRedCross:

There are two ways to enable QFA on an espo instance: 
Using Insights for large feedback volumes operations which have 1 output. This will be executed by a flowchart specifically for Insights triggering on the creation of a new insight record. These are the candidates for this: (so we remove 'code' and 'summarize per item')
- `Summarize`
- `Analyze`

Enabling the Boolean fields in individual feedback records. Note: a user can also enable these Booleans in a 'mass update' (see screenshot) which will then trigger a feedback record flowchart (not the insight flowchart). These will work for the below:

-  `Auto Coding Enabled = True`
- `Automatically flagged sensitive = True`
- `Auto Summary Enabled = True`

### LINKED ISSUE #149: Return correct coding level structure for espo consumption
The `assign-codes` endpoint should return ids of the three `CodingLevel` 1-2-3's  that were found. Changes should be made that:
- Return the ids of the items (`CodingLevel`)
- Return all coding levels assigned, not just the leaf `CodingLevel3`s

teeuwski:
> [!WARNING]
> This is a stacked PR on top of #142, please merge that one first.

Closes #116 
Closes #149 

Combines EspoCRM logic into as little seperate files as possible. 

Builds one giant `motherPayload` that contains all input for an insight to be created. API endpoints will ignore any information that is not used. 

Logic splits after the response from the API hits EspoCRM


src/qfa/api/schemas.py
```diff
@@ -479,19 +495,28 @@ class ApiCodingFramework(BaseModel):
...
         min_lengths = set(code.min_child_depth() for code in self.root_codes)
+
+        # All root codes must have identical tree depth
         if len(max_lengths.union(min_lengths)) > 1:
             raise ValueError(
                 f"All codes must have the same depth {min_lengths=} {max_lengths=}"
             )
 
+        # Require exactly 3 levels: max_child_depth of 2 means L1->L2->L3
+        if max_lengths and (next(iter(max_lengths)) != 2):
```
mariushelf: had to think about this line way too long :)
Isn't it equivalent to `if max_lengths != {2}:` ? I'd find that a lot more readable.
And you could even fold it into the condition above: `if max_lengths.union(min_lengths) != {2}:` .

src/qfa/api/schemas.py
```diff
@@ -510,15 +535,39 @@ class ApiAssignCodesRequest(ApiSingleInferenceRequestBase):
...
+    coding_level_2_name: str = Field(
+        description="Name of the selected Category level code."
+    )
+    coding_level_3_id: str = Field(
+        description="ID of the selected Code (leaf) level code."
+    )
+    coding_level_3_name: str = Field(
+        description="Name of the selected Code (leaf) level code."
+    )
+    confidence_type: float = Field(
```
mariushelf: here you use `type`, `category`, `code`, according to the UL. Above the same concept is called `coding_level_{1|2|3}`. Id suggest to align all names to the UL, and cherry-pick the addition to the UL markdown file from #159 .

src/qfa/api/routes.py
```diff
@@ -308,8 +311,12 @@ async def assign_codes(
...
         assigned_codes=[
             ApiAssignedCode(
-                code_id=assigned.code_id,
-                code_label=assigned.code_label,
+                coding_level_1_id=assigned.coding_level_1_id,
+                coding_level_1_name=assigned.coding_level_1_name,
+                coding_level_2_id=assigned.coding_level_2_id,
+                coding_level_2_name=assigned.coding_level_2_name,
+                coding_level_3_id=assigned.coding_level_3_id,
+                coding_level_3_name=assigned.coding_level_3_name,
```
mariushelf, line 320: same misalignment of names as below

src/qfa/domain/models.py
```diff
@@ -228,12 +235,26 @@ class CodingAssignmentRequestModel(BaseModel):
...
 class AssignedCodeModel(BaseModel):
-    """A single leaf code assigned to a feedback record."""
+    """A single leaf code assigned to a feedback record with full hierarchical path."""
 
     model_config = ConfigDict(frozen=True)
 
-    code_id: str = Field(description="Stable identifier from the coding framework.")
-    code_label: str = Field(description="Human-readable code name.")
+    coding_level_1_id: str = Field(description="ID of the selected Type level code.")
+    coding_level_1_name: str = Field(
```
mariushelf: same

src/qfa/services/orchestrator.py
```diff
@@ -199,8 +199,12 @@ def _build_judge_system_message(source_text: str, summary: str) -> str:
...
 @dataclass
 class _ScoredCode:
-    code_id: str
-    code_label: str
+    coding_level_1_id: str
+    coding_level_1_name: str
+    coding_level_2_id: str
+    coding_level_2_name: str
+    coding_level_3_id: str
+    coding_level_3_name: str
```
mariushelf, line 208: same

src/qfa/services/orchestrator.py
```diff
@@ -1310,8 +1314,12 @@ async def assign_codes(
 
                     candidates.append(
                         _ScoredCode(
-                            code_id=code_name,
-                            code_label=code_name,
+                            coding_level_1_id=type_node.id,
```
mariushelf: same

src/qfa/services/prompts.py
```diff
@@ -128,6 +128,7 @@ def escape_for_tag_envelope(text: str) -> str:
 def build_analyze_user_message(
     analyst_prompt: str,
     feedback_records: tuple[FeedbackRecordModel, ...],
+    output_language: str | None = None,
```
mariushelf, line 146: I this this duplicates work from #157. This here looks better though, because I just realized that #157 passes user input directly into the system prompt.

Let's keep them both for now. Follow up ticket in #161.

mariushelf (commented):
The Python parts looks technically good.

What do you think about my remark on coding level 1/2/3 vs type/category/code?

I'd advocate for adhering to the UL. If you agree to change it, then for the sake of faster iteration we can also only do the change in the user-facing API models because that has an impact on the Espo code, too, and is hard to change later. And create a follow-up ticket for the internal objects.

src/qfa/api/schemas.py
teeuwski: I don't believe that type, category or code are something real from espo. I'm asking @DGorsseRedCross if it would make sense. Otherwise I'm convinced this just vibed its' way into the application- and I'll remove any mention of these three level names.

src/qfa/services/prompts.py
teeuwski: Thanks :)

mariushelf:
Given @DGorsseRedCross reply in Teams, you can consider the Python side approved!

mariushelf (approved):
Python side LGTM!
Merging the espo-crm part is low risk, because it's not deployed immediately. So Jacopo can review even after merging.

If you agree @teeuwski , feel free to merge!

PR #142, opened 2026-06-01
feat(api): reshape schemas

### LINKED ISSUE #141: Reshape in/ output schemas
### **Explanation**
The current inference api schemas will take a lot of time to integrate into the requested EspoCRM flows. To make integrating these easier, a couple of changes are to be made before we can easily finalize the Espo integration.

### **What needs to be done**
- [x] **Split into individual / bulk endpoints**
The endpoints `summarize-aggregate` and `analyze` shall be renamed with a `-bulk` postfix. The other inference endpoints will only take a single `ApiFeedbackRecordInput` instead of a list of length > 1.
- [x] **Use standardized output for bulk responses**
The bulk responses shall all have a `pretty_output` attribute that can be used to fill in the Insight records without having diverging logic in the flow.
- [x] **Remove Anonymization flag** as this is standard and should always be used
- [ ] **Look for simplification and unification in the schemas** 

### **Acceptance Criteria**
- [x] Tested and reviewed
- [x] Documented the changes in the approriate location

teeuwski:
Closes #141

## Semantic seperation between `-bulk` and non-bulk endoints
We now have clear seperation between endpoints that process a list of feedback record and single feedback records. Some endpoints were also changed to support the assingn-to-single-record use-case better, instead of having input lists. Clear labels are given (bulk, individual) to the endpoints in fastapi docs.

## Move to `pretty_output` for bulk responses
All bulk responses now have a `pretty_output` as these are assigned in the same way to a new insight item. This simplifies the espo flow logic afterwards.

## Remove anonymization flag
As we always want to apply it


src/qfa/api/routes.py
```diff
@@ -125,17 +117,71 @@ async def analyze(
...
+            content=record.content,
+            metadata=record.metadata,
+        )
+        for record in body.feedback_records
+    )
+    domain_request = SummaryRequestModel(
+        feedback_records=feedback_records,
+        output_language=body.output_language,
+        tenant_id=tenant.tenant_id,
+    )
```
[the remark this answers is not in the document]
teeuwski: Created Issue

src/qfa/domain/models.py
```diff
@@ -35,6 +35,29 @@ class FeedbackRecordModel(BaseModel):
...
+class CodingNode(BaseModel):
+    """A node in a hierarchical coding framework."""
+
+    model_config = ConfigDict(frozen=True)
+
+    name: str = Field(description="Label for this node in the coding hierarchy.")
+    children: list["CodingNode"] = Field(
+        default_factory=list,
+        description="Child nodes. Empty for leaf-level codes.",
+    )
```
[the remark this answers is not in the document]
teeuwski: Created ticket

src/qfa/services/orchestrator.py
```diff
@@ -574,161 +541,152 @@ async def assign_codes(
...
-                                confidence_category=judge_category.score,
-                                confidence_code=judge_code.score,
-                                explanation_type=judge_type.explanation,
-                                explanation_category=judge_category.explanation,
-                                explanation_code=judge_code.explanation,
-                            )
+                    candidates.append(
+                        _ScoredCode(
+                            code_id=code_name,
+                            code_label=code_name,
```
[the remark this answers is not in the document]
teeuwski: Created ticket

mariushelf:
## Code review — overview

Reviewed the full diff and cross-checked each finding against the actual branch source (not just the patch). The reshape is clean and the mechanical rename is thorough. Summary of what this PR does:

1. **Bulk vs. per-record separation.** `/v1/analyze` → `/v1/analyze-bulk`, `/v1/summarize-aggregate` → `/v1/summarize-bulk` (tagged *Bulk Inference*). `/v1/summarize`, `/v1/assign-codes`, `/v1/detect-sensitive` become **single-record** — one `feedback_record` in, one result object out.
2. **`anonymize` flag removed everywhere** — anonymisation is now unconditional, including the judge hop.
3. **Field/type cleanups** — `FeedbackRecordModel.text` → `content`; untyped `coding_framework: dict` → typed `CodingLevels`/`CodingNode` tree; bulk responses gain an abstract `pretty_output`.

Detailed findings follow as separate comments: 2 high-severity defects, 4 medium, 3 nits, plus positives and a test-coverage note.

**Recommendation: request changes** — findings #1 and #2 are concrete defects that should be fixed before merge.

~Claude

mariushelf:
## [High] Response-only fields leaked into the per-record *request* base

`ApiSingleInferenceRequestBase` (`src/qfa/api/schemas.py:375`) declares `confidence` (`:381`) and `coding_trends` (`:390`) — both with descriptions copied verbatim from the *analyze response* ("Coverage-weighted mean of per-chunk judge faithfulness scores…"). These are output concepts.

As a result, the **request** schemas for `/v1/summarize`, `/v1/assign-codes`, and `/v1/detect-sensitive` now advertise and accept two meaningless optional fields. They default to `None` so valid requests still work, but the OpenAPI contract is wrong and misleading — a client could send `coding_trends` on a summarize request and reasonably expect it to do something.

Almost certainly a copy-paste from the response base. **Suggest removing both fields from the request base.**

~Claude

mariushelf:
## [High] `scripts/stress_analyze.py` is broken by the endpoint rename

The script still issues `POST /v1/analyze` (`scripts/stress_analyze.py:187`) — now a 404 — and `_build_body` (`:127`) emits `feedback_records` with `text` keys plus the removed `anonymize` field.

Two problems compound:
- The path no longer exists (renamed to `/v1/analyze-bulk`).
- Even if the path is fixed, `ApiFeedbackRecordInput` now requires `content` with **no `text` alias at the API layer**, so the records would 422.

`scripts/test-api.sh` was updated to `/v1/analyze-bulk`; this script was missed. Its docstring at `:127` even claims the fixture `{id, text, metadata}` shape "already matches the schema" — no longer true.

~Claude

mariushelf:
## [Medium] Dead code: `Orchestrator._align_record_items`

`_align_record_items` (`src/qfa/services/orchestrator.py:1194`) now has zero callers on this branch — it's the only reference left. It existed to reconcile LLM output count against input records for the old multi-record `summarize`; with single-record endpoints it's orphaned.

Suggest removing it.

~Claude

mariushelf:
## [Medium] `code_id` collapses into the label

`assign_codes` now sets `code_id=code_name` (was `str(code.get("code_id", ""))`) because `CodingNode` carries only `name`/`children` — no stable id.

Consequences:
- Any consumer keying on `code_id` as an identifier distinct from the human-readable label breaks.
- Two codes sharing a label anywhere in the tree become indistinguishable by id.

Please confirm label-as-id is acceptable, or add an `id` / `code_id` field to `CodingNode` so the assigned `code_id` stays distinct from `code_label`.

~Claude

mariushelf:
## [Medium] Inconsistent `text` → `content` back-compat

The domain `FeedbackRecordModel` gained `validation_alias=AliasChoices("content", "text")` (`src/qfa/domain/models.py`), but the API edge (`ApiFeedbackRecordInput`) renamed `text` → `content` with **no** alias.

So external clients still sending `text` get a 422 — the domain alias only helps internal construction (e.g. corpus fixtures), never the public API. The back-compat is therefore half-applied and a bit misleading.

Pick one policy and be consistent: either accept `text` at the API edge too, or drop the domain-level alias.

~Claude

mariushelf:
## [Medium] EspoCRM payload scripts may be stale

`docs/integrations/espo-crm.md` now maps `set_summarize_*` → `/v1/summarize`, but `/v1/summarize` changed from many-records to a single `feedback_record`. The server-side payload builders under `scripts/espo_crm/` (e.g. `set_summarize_per_item_payload`, `set_summarize_aggregate_payload`) are **not** part of this PR.

Since the stated goal is to "simplify the espo flow logic afterwards", please confirm these scripts are updated elsewhere — otherwise the integration will break against the new single-record / renamed shapes.

~Claude

mariushelf:
## [Nit] Inert request surface on the bulk base

`ApiBulkInferenceRequestBase` exposes `mode` + `output_language`, but:
- the `analyze-bulk` route never forwards `output_language` to the domain request, and
- `summarize-bulk` can't set `prompt`, even though the orchestrator's `summarize_bulk` still reads `request.prompt`.

So part of the request surface is accepted but ignored. Suggest pruning fields the handlers don't consume so the schema matches behaviour.

~Claude

mariushelf:
## [Nit] Grammar in a comment

`src/qfa/api/schemas.py`:

```python
# note: no response base model for since these are all different shapes
```

"for since" reads broken — probably meant "no response base model **here**, since these are all different shapes".

~Claude

mariushelf:
## [Nit] `SummaryRequestModel` retains fields the bulk request can't set

`SummaryRequestModel` keeps `prompt` / `output_language`, used only by the bulk orchestrator path, while the bulk **request** schema (`ApiSummarizeBulkRequest`) no longer accepts `prompt`. The result is a domain field that can never be populated from the API. Slightly confusing — consider trimming.

~Claude

mariushelf:
## Positives

A few things worth calling out that are well done:

- **Unconditional anonymisation, including the judge hop**, is a genuine security improvement. The retained test `test_judge_call_does_not_see_raw_analyst_prompt_when_anonymized` still guards the second-hop PII leak, and dropping the `anonymize=False` escape hatch removes a foot-gun.
- The **abstract `pretty_output` computed-field** on the bulk response base cleanly forces every bulk response to implement it — nice use of the ABC.
- **Test updates are thorough and retain docstrings** (matches project convention): single-record bodies, the `text`→`content` renames, and the `summarize-bulk` operation mapping are all covered.

~Claude

mariushelf:
## Test-coverage note & verification

**Coverage gap:** no test asserts that single-inference *responses* omit `confidence`/`coding_trends`, nor that those fields are absent from the single-inference *requests* — which is exactly why finding #1 wasn't caught. A small schema assertion would lock the contract.

**Verification done for this review:** read the full diff and cross-checked each finding against the actual branch source via `git grep` / `git show`, not just the patch text. I did **not** execute the test suite (read-only review from a different branch; the e2e/embedding tests pull external resources). Findings #1, #3, #5 are static-certain; #4 and #6 warrant author confirmation.

~Claude

teeuwski:
@mariushelf thank you for the review. The ❤️ is implemented, otherwise:

> [Medium] code_id collapses into the label

Already an Issue

> [Medium] EspoCRM payload scripts may be stale

Picking up in #137

> [Nit] Inert request surface on the bulk base

Created #152 

> [Nit] SummaryRequestModel retains fields the bulk request can't set

Created #153

> Positives

Thank you for the complements claude code.

scripts/test-api.sh
```diff
@@ -14,8 +14,8 @@ echo "=== Health check ==="
 http GET "${BASE_URL}/v1/health"
 
 echo ""
-echo "=== Analyze ==="
-http POST "${BASE_URL}/v1/analyze" \
+echo "=== Analyze Bulk ==="
+http POST "${BASE_URL}/v1/analyze-bulk" \
   "Authorization:Bearer ${QFA_API_KEY}" \
   documents:='[
```
mariushelf: documents should be records everywhere. Maybe create a follow-up ticket if documents is still used in the API.

src/qfa/api/schemas.py
```diff
@@ -242,183 +306,129 @@ class ApiAnalyzeResponse(BaseModel):
...
-    summaries: list[ApiFeedbackRecordSummary] = Field(
-        description="Title and summary for each submitted feedback record.",
-    )
-    used_anonymization: bool = Field(
-        description="Indicates whether anonymization was applied to the feedback text.",
+
+class ApiSingleInferenceRequestBase(BaseModel, ABC):
+    """Base request for inference endpoints that return per-feedback-record outputs."""
+
+    feedback_record: ApiFeedbackRecordInput = Field(
```
mariushelf: So this changes the shape of the API significantly.

Previous we had endpoints
1. that received a list of records, and ran inference for each of them, and returned a result for each of them.
2. that received a list of records and returned a single result inferred from all of them

Now, the first endpoint only receives and infers a single record.
That's a valid choice, however it moves logic and optimisations from the backend to the caller.

Before, we could start several parallel inference tasks.
Now, we only receive a single record, and parallelization must happen at the caller of the REST API.

Personally, I would keep the list-in, list-out shape and let the backend do the work.

src/qfa/api/schemas.py
```diff
@@ -92,34 +131,71 @@ def _assign_codes_request_examples() -> list[dict[str, Any]]:
...
+class ApiBulkInferenceRequestBase(BaseModel, ABC):
+    """Base request for inference endpoints that process bulk feedback records."""
+
+    feedback_records: list[ApiFeedbackRecordInput] = Field(
+        min_length=1,
+        description="Non-empty list of feedback records to process.",
+    )
+
+    mode: Literal["single_pass"] = Field(
+        default="single_pass",
```
mariushelf: I think the `mode` is specific to the endpoint. Also, if `analyze` support "hierarchical" but "summarize-bulk" does not, but the `mode` allows both, then Interface Segregation Principle in SOLID is violated (subclasses should not depend on interfaces that they don't use, or, here, constants)

src/qfa/api/schemas.py
```diff
@@ -242,183 +306,129 @@ class ApiAnalyzeResponse(BaseModel):
...
-    """Response body for the ``POST /v1/summarize`` endpoint."""
+##### Per-feedback-record requests #####
 
-    summaries: list[ApiFeedbackRecordSummary] = Field(
-        description="Title and summary for each submitted feedback record.",
-    )
-    used_anonymization: bool = Field(
-        description="Indicates whether anonymization was applied to the feedback text.",
+
+class ApiSingleInferenceRequestBase(BaseModel, ABC):
```
mariushelf: for discovery, I'd move the ABCs to the top of the module.

src/qfa/domain/models.py
```diff
@@ -36,6 +36,29 @@ class FeedbackRecordModel(BaseModel):
...
+
+
+class CodingLevels(BaseModel):
+    """A tree of coding nodes defining the full hierarchical coding framework."""
```
mariushelf, line 52: Maybe we could call this "CodingFramework" then to match the docstring and the UL

src/qfa/domain/models.py
```diff
@@ -125,6 +148,17 @@ class SummaryRequestModel(BaseModel):
     tenant_id: str = Field(description="Tenant identifier injected by the auth layer.")
 
 
+class SingleSummaryRequestModel(BaseModel):
+    """A request to summarize a single feedback record."""
```
mariushelf, line 152: in theory we could also have a base class or mixin to inject the feedback_record into every model, right? Though this is not very verbose and I'm happy with it. Just noticing an asymmetry :)

src/qfa/services/orchestrator.py
```diff
@@ -310,7 +313,7 @@ def _is_retained_analyze_placeholder(cls, placeholder: str) -> bool:
             for entity_type in cls._ANALYZE_RETAINED_PLACEHOLDER_TYPES
         )
 
-    async def analyze(
+    async def analyze_bulk(
```
mariushelf, line 316: I am not sure whether "bulk" as the best / most concise choice of words.
Bulk could also mean "analyze many at records once, return one result per record". And that is an option I'd suggest to keep :)

From the top of my head I don't have a better word. "aggregate" describes it well, but is not nice to read. "reduce" maybe too, but still does not flow.

Let's brainstorm?

mariushelf (commented):
Technically sound.

I'll repeat 2 inline comments here so we can have a centralised discussion.

## API shape and impact on optimisations
This PR changes the shape of the API significantly:

Previous we had endpoints

1. that received a list of records, and ran inference for each of them, and returned a result for each of them.
2. that received a list of records and returned a single result inferred from all of them

Now, the first endpoint only receives and infers a single record.

That's a valid choice, however it moves logic and optimisations from the backend to the caller.

Before, we could start several parallel inference tasks, or pass all documents at once to the LLM.

Now, we only receive a single record, and parallelization must happen at the caller of the REST API, or we take into account longer total inference time.

Personally, I would keep the list-in, list-out shape and let the backend do the work.

## Naming

I am not sure whether "bulk" as the best / most concise choice of words.
Bulk could also mean "analyze many at records once, return one result per record". And that is an option I'd suggest to keep :)

From the top of my head I don't have a better word. "aggregate" describes it well, but is not nice to read. "reduce" maybe too, but still does not flow.

If we decide to scrap the list-in/list-out shape forever, than "bulk" is ok. 


Let's brainstorm?

src/qfa/api/schemas.py
```diff
@@ -92,34 +131,71 @@ def _assign_codes_request_examples() -> list[dict[str, Any]]:
...
+    mode: Literal["single_pass"] = Field(
+        default="single_pass",
+        description=(
+            "Analysis mode. ``single_pass`` is the only supported value in this"
+            " version; additional modes may be added in future versions."
+        ),
+    )
+
+    output_language: str | None = Field(
+        default=None,
```
mariushelf, line 170: same here.. not all endpoints create unstructured output. though this one feels less off than `mode` on all endpoints. What do you think?

teeuwski: I think all bulk requests *should* do so, but maybe this is not implemented (pretty sure it is not)

src/qfa/domain/models.py
teeuwski: Yeah, totally agree. Could be interesting in the future

teeuwski:
> Technically sound.
> 
> I'll repeat 2 inline comments here so we can have a centralised discussion.
> 
> ## API shape and impact on optimisations
> This PR changes the shape of the API significantly:
> 
> Previous we had endpoints
> 
> 1. that received a list of records, and ran inference for each of them, and returned a result for each of them.
> 2. that received a list of records and returned a single result inferred from all of them
> 
> Now, the first endpoint only receives and infers a single record.
> 
> That's a valid choice, however it moves logic and optimisations from the backend to the caller.
> 
> Before, we could start several parallel inference tasks, or pass all documents at once to the LLM.
> 
> Now, we only receive a single record, and parallelization must happen at the caller of the REST API, or we take into account longer total inference time.
> 
> Personally, I would keep the list-in, list-out shape and let the backend do the work.

The EspoCrm triggers workflows per-created `Insight` with a set of feedbacks selected (when we use `-bulk`), and otherwise per saved `Feedback` item. Right now, this makes for VERY simple flowcharts compared to the alternative.

> ## Naming
> I am not sure whether "bulk" as the best / most concise choice of words. Bulk could also mean "analyze many at records once, return one result per record". And that is an option I'd suggest to keep :)
> 
> From the top of my head I don't have a better word. "aggregate" describes it well, but is not nice to read. "reduce" maybe too, but still does not flow.
> 

My vibe is that list-in-and-out is dead for the reasons described above, we could discuss with Daan - but I'd like to merge. Let's change the name in case it makes sense?

mariushelf (approved):
Ok, let's go!

If our Espo as our only consumer prefers this API shape, then that's the go :)

LGTM!

PR #146, opened 2026-06-03
feat(anonymizer): silence non-PII spaCy NER labels in Presidio logs

mariushelf:
Silence non-PII spaCy NER labels in Presidio anonymizer logs (`presidio_anonymizer.py`), so the logs surface only PII-relevant detections instead of every NER label.

Extracted as a standalone change from the #144 embedding-model branch, where it was unrelated scope. Self-contained: +23 lines in one file, no dependence on the hierarchical-analysis stack.


teeuwski (approved):
LGTM, thanks!

PR #147, opened 2026-06-05
fix(api): tolerate empty feedback content without failing requests (#138)

### LINKED ISSUE #138: [BUG]  Handle records with empty feedbackDescription in both summarize endpoitns to prevent silent 422 failures. Now EspoCRM will fail is any records has empty description.
### **Who notified us about the bug?**
Paul van Houtum

### **Describe the bug**
Handle records with empty feedbackDescription in both summarize endpoitns to prevent silent 422 failures. Now EspoCRM will fail is any records has empty description. 

### **To Reproduce**
EspoCRM insights -> any summarize endpoint -> add empty feedback record -> no result

PaulvH63:
Fixes #138.

## Problem

EspoCRM submits feedback records with blank descriptions. The API rejected the **entire** request with a 422 that EspoCRM swallowed, so a single blank record silently broke a whole batch — no result was written back.

## Approach

The messy-input tolerance lives in the API driving adapter, keeping the domain `FeedbackRecordModel.content` invariant intact (this differs from the first revision of this PR, which relaxed the domain too):

- `ApiFeedbackRecordInput.content` relaxed to `min_length=0` so empty content no longer 422s (shared across all endpoints).
- **Bulk** endpoints (`analyze-bulk`, `summarize-bulk`) drop empty-content records (logged) and process the rest; an all-empty batch returns a 200 empty result.
- **Per-record** endpoints (`summarize`, `assign-codes`, `detect-sensitive`) short-circuit to a 200 empty result with no LLM call when content is empty.

Dropping never misaligns EspoCRM results: bulk endpoints return one aggregate output and per-record endpoints echo the source `id`, so results are matched by id, never by position.

## Tests

New `TestEmptyFeedbackContent` covers mixed and all-empty batches across every record-accepting endpoint (each fails with a 422 before the fix, passes after). The domain stays strict — `test_empty_text_raises` is unchanged and still green. Full `make test` (429 passed) + `make lint` + import-linter contracts all green.

## Notes

Rebased onto current `main` after the #142 schema reshape. The original two commits could not be preserved (the `text` field was renamed to `content` and `ApiSummarizeFeedbackRecord` was removed), so the fix is re-implemented as a single commit on top of `main`.


teeuwski:
@mariushelf we need to check if the original issue is still relevant

mariushelf:
I think it is still relevant. The `FeedbackRecordModel` domain model expects a min content size of 1. So any empty records that the user passes to the bulk api will fail the domain validation. This PR works around that by dropping them.

teeuwski (approved):
LGTM

PR #156, opened 2026-06-09
feat(api): translate pretty_output headers to output_language (#155)

### LINKED ISSUE #155: translate headers in pretty_output
Output headers (QUALITY, TITLE, SUMMARY) are hardcoded English terms. 
Should be translated to he `output_language`.

mariushelf:
Closes #155.

## Problem

The human-readable `pretty_output` block (returned by `/v1/summarize` and
`/v1/summarize-aggregate` and rendered straight into an EspoCRM field) used
hardcoded English headers (`QUALITY:`, `TITLE:`, `SUMMARY:`). When a caller
requested another `output_language`, the title/summary *values* came back
translated but the *labels* stayed English — a mixed-language block.

## Approach

Pure backend change (no EspoCRM script edits). The headers are now localized to
the request's `output_language` via a static table for seven languages — the
five official languages of the Red Cross/Red Crescent Movement (English,
French, Spanish, Arabic, Russian) plus Dutch and Ukrainian — with **English
fallback** for any other or absent value.

- `_HEADER_LABELS` + `_resolve_language` in `qfa.api.schemas` (case-insensitive
  match on language name or ISO 639-1 code; falls back to `en`).
- `_create_pretty_output` gains a `language` param; the three headers are
  resolved from the table. Each label is padded to a fixed column so the
  **English output is byte-for-byte unchanged**.
- Both response models carry `output_language` as an **excluded** field
  (`exclude=True`) — a render input only, kept out of the serialized JSON so the
  data contract is not polluted. The existing `pretty_output` computed field
  passes it through.
- The two routes pass `body.output_language` at construction.
- Technical `Feedback-ID:` / `IDs:` labels are intentionally **not** localized.

No LLM translation (deferred, YAGNI) — the label vocabulary is tiny and fixed, so
a static table is deterministic, free, and human-reviewable. A cached
translate-once-per-language LLM fallback for arbitrary languages can be added
later if needed.

## Notes

- The non-English translations are a first draft and should get a
  native-speaker review before release (humanitarian context).
- Accepted cosmetic limitations: column alignment may shift for longer
  translated labels; Arabic renders RTL inside the LTR ASCII block.
- Also moves a mis-placed mid-file `TrendPeriod` import to the top of
  `schemas.py` (pre-existing E402 that blocked lint).

## Testing

- `make test` → 455 passed; `make lint` → clean, all 3 import-linter contracts kept.
- New unit tests: language resolver (names + ISO codes + fallback), header
  translation (per-record and aggregate paths), English regression guard,
  serialization-exclusion for both models.
- New route-level tests for `/v1/summarize` and `/v1/summarize-aggregate`
  localization.
- Coverage was broadened by a multi-agent review→verify→fix pass.

### DIFF
--- a/src/qfa/api/schemas.py
+++ b/src/qfa/api/schemas.py
@@ -5,6 +5,7 @@ HTTP contract can evolve independently of the core domain.
 """
 
 import json
+import logging
 from abc import ABC, abstractmethod
 from pathlib import Path
 from typing import Any, Literal, override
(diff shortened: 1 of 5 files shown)


src/qfa/api/schemas.py
```diff
@@ -11,8 +11,77 @@
...
+
 _SEPARATOR = "------------------------------------------------------------"
 
+# Localized headers for the human-readable ``pretty_output`` block (#155).
+# Keyed by label, then by ISO 639-1 code. The seven supported languages are the
+# five official languages of the International Red Cross and Red Crescent
+# Movement (en, fr, es, ar, ru) plus Dutch (nl) and Ukrainian (uk). English is
+# the fallback for any unsupported or absent ``output_language``.
+# The non-English translations are pending native-speaker review.
+_HEADER_LABELS: dict[str, dict[str, str]] = {
```
mariushelf, line 27: this is auto-generated by Claude. 
I have no idea if anything but English and Dutch are correct.
Who can check that?

src/qfa/api/schemas.py
```diff
@@ -37,13 +104,22 @@ def _create_pretty_output(
...
+    language: str | None = None,
 ) -> str:
     """Build a human-readable text block for display in EspoCRM.
 
     All arguments are optional because this function is shared across endpoints;
     each endpoint passes only the fields relevant to it. Omitted fields are
     excluded from the output.
+
+    The ``QUALITY``/``TITLE``/``SUMMARY`` headers are localized to ``language``
+    (an ``output_language`` value), falling back to English (#155). The
```
mariushelf: remove reference to PRs/issues

src/qfa/api/schemas.py
```diff
@@ -15,6 +15,73 @@
...
+
+def _resolve_language(output_language: str | None) -> str:
+    """Map a free-text ``output_language`` to a supported ISO code.
+
+    Matches English language names and ISO 639-1 codes case-insensitively;
+    returns ``"en"`` for ``None`` or any unsupported value.
+    """
+    if not output_language:
+        return "en"
+    return _LANGUAGE_ALIASES.get(output_language.strip().lower(), "en")
```
mariushelf: add logging if requested output_language does not exist and we fallback to "en"

mariushelf: Done — `_resolve_language` now logs a WARNING (echoing the requested value) before falling back to English when a present-but-unsupported language is requested. Absent/empty and recognized values stay quiet. Added two tests covering both the warn and no-warn paths.

~Claude

src/qfa/api/schemas.py
mariushelf: Done — removed the `(#155)` references from the code comments, docstrings, and test docstrings (kept it only in the commit subject, which follows the repo's existing convention). The pre-existing `(#124)` on the analyze `period` field is untouched as it's not part of this change.

~Claude

src/qfa/api/schemas.py
mariushelf: Agreed — only `en` and `nl` are trusted; `fr/es/ar/ru/uk` are AI-generated and unverified. Per discussion we're keeping all seven with the existing `# The non-English translations are pending native-speaker review.` caveat in the table, and unsupported/absent requests already fall back to English. **These five still need native-speaker sign-off before we rely on them** — that needs a human reviewer (ideally Movement-language speakers); I can't verify them. Happy to open a tracking issue or gate them to en+nl if you'd prefer not to ship unverified strings.

~Claude

teeuwski (approved):
LGTM. Happy that the fallback is `en`

Issue #139, opened 2026-05-29
[BUG]  output_language field has no effect — prompt ignores it and pretty_output labels are hardcoded    ▎ English.


PaulvH63:
### **Who notified us about the bug?**
Paul van Houtum

### **Describe the bug**
The output_language field is sent in the payload and accepted by the API, but the LLM prompt does not instruct the model to respond in that language — so the model defaults to the language of the input text. Additionally, the pretty_output labels (QUALITY, TITLE, SUMMARY) are hardcoded English strings in _create_pretty_output in schemas.py, so they never change regardless of what output_language is set to.

### **To Reproduce**
1. EspoCRM choose any summarize endpoint
2. alter output lanaguage 
3. observe that nothgin happens

PR #157, opened 2026-06-10
feat(analyze): pass output_language through to the LLM (#154)

### LINKED ISSUE #154: pass `output_language` through to LLM
pass `output_language` through to LLM in all endpoints which use it.
See parent issue #139 for details.

mariushelf:
## What & why

Closes #154 (sub-task of #139).

The `analyze-bulk` endpoint accepted `output_language` via the shared
`ApiBulkInferenceRequestBase`, but the route never forwarded it to the
domain request and the analyze prompts never mentioned language. The model
therefore answered in the language of the input records regardless of what
was requested. `summarize-bulk` already wired this up correctly; this PR
brings `analyze-bulk` to parity.

## Changes

- **Domain** — add `output_language: str | None` to `AnalysisRequestModel`.
- **Route** — `analyze_bulk` now forwards `body.output_language`.
- **Prompts** — new `build_output_language_instruction()` helper returns a
  system-prompt suffix (`"Write the analysis in <language>."`) or `""` when
  no language is requested, so callers append unconditionally.
- **Single-pass** — the directive is appended to the analyst system message.
- **Hierarchical** — the directive is applied at the **reduce** (synthesis)
  step, which produces the final user-facing analysis, and threaded through
  the tree-reduce recursion. Token budgeting accounts for it because the
  augmented system message is what `_fits`/`_group_partials_to_budget` see.

The directive always lives in the **system** message, never the untrusted
user message, so a feedback record cannot spoof or override the language.

## Scope decision

`output_language` controls the **primary analysis output** only. Two things
are intentionally left in their default language:

- The **map** step's partial analyses — intermediate, never returned.
- The judge's **`uncertainty_explanation`** — QA meta-text; matching the
  existing summarize precedent, which does not translate judge output.

Easy to extend if reviewers want the judge explanation localized too.

## Tests (TDD, all red first)

- `build_output_language_instruction`: suffix when set, empty when `None`/`""`.
- Orchestrator single-pass: directive reaches the analyst (first) call;
  absent when unset.
- Orchestrator hierarchical: directive reaches the top-level reduce prompt.
- Route: `output_language` from the body reaches `AnalysisRequestModel`
  (and defaults to `None` when omitted).

`make test` (432 passed), `make lint` (ruff + ty + import-linter contracts)
all green.

## Docs

Added an `output_language` row to the analyze-bulk request field reference in
`docs/rest-api/index.md`.


teeuwski (approved):
LGTM, really important for end users!

PR #159, opened 2026-06-10
feat(assign-codes)!: return full coding hierarchy, drop code_id (#149)

### LINKED ISSUE #149: Return correct coding level structure for espo consumption

mariushelf:
> [!WARNING]
> **Breaking API change — the `/v1/assign-codes` response shape changes.**
> - **`code_id` is removed** from each assigned code. It only ever mirrored `code_label` (the inbound `coding_levels` nodes carry names, not stable ids), so it carried no information.
> - **Two fields are added** to each assigned code: `type_label` (level 1) and `category_label` (level 2), alongside the existing `code_label` (level 3 leaf).
>
> Any EspoCRM script or client parsing the assign-codes response must stop reading `code_id` and read the three `*_label` fields. EspoCRM maps the upper two levels back to its `CCodingLevel1/2` records **by name**.

Closes #149.

## What & why

The `assign-codes` endpoint previously returned only the leaf code per assignment. Issue #149 asks it to return **all three coding levels** so EspoCRM can map an assignment back to `CCodingLevel1/2/3`.

Since `ApiCodingNode` carries only names (no ids), there are no stable Type/Category ids to return — so the assignment is made self-describing by **name** instead:

- `type_label` — Type level (1)
- `category_label` — Category level (2)
- `code_label` — Code level (3, leaf)

`code_id` is dropped because it was a duplicate of `code_label`.

## Changes

- **Domain** (`AssignedCodeModel`): add `type_label` / `category_label`, drop `code_id`.
- **Orchestrator**: thread the Type and Category names (already in scope in the pick loop) through `_ScoredCode` into the assignment.
- **API** (`ApiAssignedCode` + route): expose the two label fields, drop `code_id`, add OpenAPI field descriptions; expand the route docstring with response semantics.
- **Docs** (`ubiquitous_language.md`): define *Coding type* / *Coding category* / *Code* inline by the *Coding framework* entry.

## Tests

- Service-level (`tests/services/test_orchestrator.py`): drives the real `assign_codes` over a 3-level framework and asserts the assignment carries all three labels.
- API-level (`tests/api/test_routes.py`): asserts the response exposes the three labels and **no longer** includes `code_id`.

Full suite: **426 passed**; `make lint` clean (3 hexagonal contracts kept); `make docs` builds.

### DIFF
--- a/docs/ubiquitous_language.md
+++ b/docs/ubiquitous_language.md
@@ -19,7 +19,10 @@ name in all contexts such as code, documentation, dashboards and training materi
 | Feedback metadata | Contextual attributes that describe a feedback record for governance, analysis, and responsibility, including operational and ethical flags. | Extra fields, system data |
 | Feedback dataset | Analytical abstraction that supports aggregation, filtering, and ML workflows across storage backends; may be derived or time‑bound. | Database, table |
 | Coding | Standard qualitative analysis practice emphasising interpretation; a single feedback record may be coded multiple times. | Tagging only |
-| Coding framework | Flexible, evolving structure for organising codes; supports iteration and learning over fixed classification. | Taxonomy, list of tags |
+| Coding framework | Flexible, evolving structure for organising codes; supports iteration and learning over fixed classification. A framework is a three-level hierarchy of **coding type → coding category → code**, and every assignment names all three levels. | Taxonomy, list of tags |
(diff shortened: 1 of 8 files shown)


teeuwski:
Implemented in #137

PR #162, opened 2026-06-11
feat(api): unify and harden output_language passed to the LLM (#161)

### LINKED ISSUE #161: unify and harden passing output language to LLM
The work done in #157 passes the output language unescaped and unchecked into the system prompt.

#137 duplicates some of its work in a more secure way. 

We need to cherry-pick the best of both implementations and remove redundant and insecure code.

mariushelf:
## What & why

Closes #161.

`output_language` was reaching the LLM two different ways: #157 (merged) interpolated it **unescaped and unchecked** into the system prompt, and #137 added a separate, escaped copy in the *user* message. This PR unifies on one hardened path and removes the redundant/insecure handling.

## Approach (agreed with @mariushelf)

**Sanitize-and-pass-through, no allow-list.** `output_language` is free text — any language the model can produce — so we never reject it or map it to a code. We only neutralize it at the API boundary, then pass it through.

- **`sanitize_output_language()`** (in `qfa.api.schemas`, next to `_resolve_language`): NFC-normalize, collapse all internal whitespace to single spaces, keep only Unicode **letters and combining marks** (any script) plus space / hyphen / parentheses / apostrophe, drop everything else (`.` `:` digits braces backticks control chars), cap at 50 chars, never reject. `None`/empty/all-junk → `None`.
  - Permits genuine names: `Brazilian Portuguese`, `Chinese (Simplified)`, `हिन्दी`, `O'odham`.
  - Neutralizes sentence-injection: `English.\nIgnore previous instructions` → `English Ignore previous instructions` (an inert fragment).
- **Boundary enforcement**: a `field_validator` on `ApiBulkInferenceRequestBase` runs the sanitizer, so the **domain receives an already-trusted value** and the services layer never interpolates raw client input.
- **One directive builder**: `build_output_language_instruction(value, subject="analysis")` is the single source; the inline summarize-aggregate f-string is replaced with `subject="title and summary"`. The directive lives only in the **system message**.

The fixed **7-language `pretty_output` header table** (`_resolve_language` / `_HEADER_LABELS`) is a separate concern and is left untouched.

## Reconciled with #137 (now merged) ✅

Rebased onto `main` after #137 landed. #137 had added a redundant `<output_language>` envelope in the analyse **user** message — duplicating the system-message directive and mixing a config field into the untrusted, anonymized record envelope. This PR **removes** that path: the `output_language` parameter on `build_analyze_user_message`, the envelope itself, and the orchestrator call that threaded it through. A regression test asserts the analyse user message never carries `output_language`. (This removal is the "remove redundant and insecure code" half of #161.)

## Tests (TDD, red first)

- Sanitizer: real names incl. non-Latin scripts and decomposed-Latin accents; whitespace collapse; punctuation/digit/control-char stripping; injection → inert fragment; 50-char cap (no trailing space); empty/all-junk → `None`.
- Boundary validator: messy body value arrives sanitized; clean value unchanged; omitted → `None`.
- Builder: default vs custom subject; builder does not itself sanitize.
- Orchestrator: directive in the system message; **absent from the user message**; absent entirely when unset.

`make test` → **499 passed** (42 LLM/e2e deselected). `make lint` → import-linter 3/3 contracts kept.


teeuwski (dismissed):
Looks good to me! 🏴‍☠️🏴‍☠️🏴‍☠️🏴‍☠️
---------------
Arrr, matey! Let me be tellin' ye what be found in these here feedback records.
**Main Themes:**
**Water Distribution:** 
The folks be mentionin' that the water distribution be well organized, but there be a bit of a wait, like standin' in line for three hours.
**Medical Services:** The medical staff be praised for their professionalism, but there be a grumble about not havin' enough medicine.
**Sentiment:**
- The sentiment be mixed, with both positive and negative feedback. 
-  Positive: The organization of water distribution and the professionalism of the medical staff be makin' the folks happy.
- Negative: The long wait for water and the lack of medicine be causin' some frustration.

So, in summary, the main themes be water distribution and medical services, with a mix of happy and frustrated feelings among the folks. Savvy?

teeuwski (approved):

PR #186, opened 2026-06-17
fix: formalise feedback record metadata contract

### LINKED ISSUE #143: Formalise the feedback-record metadata contract
## Context

While running the corpus-quality notebook, `coding_trends` came back empty
("No coding_trends produced for this sample") even though every record carries a
date. Root cause: a silent field-name mismatch — the corpus stored the record
date under `creation_date`, but `ANALYZE_CODING_TREND_DATE_FIELD` defaults to
`created`. `build_coding_trend_table` looks up the configured key, finds nothing
parseable, and returns `None` **by design** (best-effort omission). The same
setting also drives the within-chunk date sort used by the map-step chunking, so
that was silently a no-op too.

The immediate symptom is already remediated on `feat/corpus-runner` (the corpus
and `scripts/generate_corpus.py` now both use `created`). This ticket is the
**structural** follow-up: the feedback-record metadata is an unenforced,
free-form `dict[str, str | int | float]`, and the pipeline silently depends on
specific keys via settings defaults.

## Problem

- **No schema.** Nothing validates which metadata fields exist, their names, or
  their types.
- **Implicit contract.** Several consumers depend on field names by convention
  only: `ANALYZE_CODING_TREND_DATE_FIELD` (`created`),
  `ANALYZE_CODING_TREND_CODE_FIELDS` (`codes`), the date-ordered chunking,
  `sensitive`, `year`, `region`, …
- **Drift fails silently.** A renamed or missing field degrades to an empty
  trend table and unsorted chunks instead of a loud error — exactly the failure
  we just hit. Three failure surfaces hide behind one informal dict.

## Proposed scope

- [ ] Define the metadata contract — a typed model (pydantic) or a documented
      schema naming the fields the pipeline consumes, their types, and
      required-vs-optional status.
- [ ] Single source of truth for field names so generator ↔ corpus ↔ settings
      defaults cannot drift apart again.
- [ ] Fail loud, not silent: emit a WARNING (or a validation error where
      appropriate) when a configured field (date / code) matches zero records,
      so a future mismatch surfaces immediately instead of as an empty result.
- [ ] Document the contract under `docs/` (per the keep-docs-in-sync rule).

## Notes

Relates to the hierarchical-analysis coding-trend table and the date-ordered
map-step chunking introduced on `feat/corpus-runner`.

olafdegraaff:
Non-AI comment: "I am still understanding the structure of the code, so please be hars on me if anything is in the wrong place" ;) 

Closes #143 

## Summary

- Add `METADATA_DATE_FIELD` and `METADATA_CODE_FIELD` constants to `settings.py` as the single source of truth for pipeline-consumed metadata field names, so the corpus generator, pipeline, and settings defaults cannot drift apart again
- Add `FeedbackRecordMetadata` to `qfa.domain.models` as a documented schema of the two fields the pipeline consumes (`created`, `codes`), their types, and optional status
- Emit a `WARNING` in `build_coding_trend_table` when the configured date field or code fields match zero records, so a field-name mismatch surfaces in logs immediately instead of silently returning an empty trend table
- Document the metadata contract in `docs/architecture/05-data-model.md`

## Test plan

- [ ] `uv run pytest tests/services/test_coding_trends.py tests/services/test_clustering.py tests/test_large_corpus.py tests/test_analyze_corpus.py tests/services/test_orchestrator.py tests/services/test_orchestrator_hierarchical.py -q`
- [ ] Confirm a WARNING log appears when `date_field` matches no records

### DIFF
--- a/scripts/generate_corpus.py
+++ b/scripts/generate_corpus.py
@@ -61,6 +61,8 @@ import matplotlib.pyplot as plt
 import numpy as np
 import yaml
 
+from qfa.settings import METADATA_CODE_FIELD, METADATA_DATE_FIELD
+
 logger = logging.getLogger("generate_corpus")
 
 # --- Timeline -----------------------------------------------------------------
@@ -197,7 +199,7 @@ PATTERN_PRIORITY: list[str] = [
 
 def _record_leaves(record: dict[str, Any]) -> list[str]:
     """Return the list of leaf code strings carried by a record."""
-    raw = record.get("metadata", {}).get("codes", "") or ""
+    raw = record.get("metadata", {}).get(METADATA_CODE_FIELD, "") or ""
     return [c.strip() for c in raw.split(",") if c.strip()]
 
 
@@ -380,7 +382,7 @@ def assign_dates(
             )
         for idx, date in zip(indices, dates, strict=True):
             iso = date.isoformat()
-            records[idx]["metadata"]["created"] = iso
+            records[idx]["metadata"][METADATA_DATE_FIELD] = iso
             records[idx]["metadata"]["year"] = date.year
 
     return dict(by_pattern)
@@ -495,7 +497,9 @@ def make_plot(
                 for idx in indices:
                     if leaf not in _record_leaves(records[idx]):
                         continue
-                    d = dt.date.fromisoformat(records[idx]["metadata"]["created"])
+                    d = dt.date.fromisoformat(
+                        records[idx]["metadata"][METADATA_DATE_FIELD]
+                    )
                     m = (d.year - START.year) * 12 + (d.month - START.month)
                     counts[m] += 1
                 ax.bar(
@@ -510,7 +514,7 @@ def make_plot(
 
             totals = np.zeros(N_MONTHS, dtype=int)
             for idx in indices:
-                d = dt.date.fromisoformat(records[idx]["metadata"]["created"])
+                d = dt.date.fromisoformat(records[idx]["metadata"][METADATA_DATE_FIELD])
                 m = (d.year - START.year) * 12 + (d.month - START.month)
                 totals[m] += 1
             ax.bar(
@@ -525,7 +529,7 @@ def make_plot(
             ax.legend(loc="upper left", fontsize=7, ncols=2, frameon=False)
         else:
             dates = [
-                dt.date.fromisoformat(records[i]["metadata"]["created"])
+                dt.date.fromisoformat(records[i]["metadata"][METADATA_DATE_FIELD])
                 for i in indices
             ]
             days = (
@@ -775,7 +779,7 @@ def _sample_metadata(
         "source": rng.choices(sources, weights=source_weights, k=1)[0],
         "year": START.year,
         "sensitive": sensitive,
-        "codes": leaf.code_id,
+        METADATA_CODE_FIELD: leaf.code_id,
     }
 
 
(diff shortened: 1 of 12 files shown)


src/qfa/domain/models.py
```diff
@@ -32,7 +46,11 @@ class FeedbackRecordModel(BaseModel):
     )
     metadata: dict[str, str | int | float | bool] = Field(
```
teeuwski, line 33: Can't this just be `metadata: FeedbackRecordMetadata`?

teeuwski (commented):
I feel like a basic implementation of a pydantic model would work better here, instead of this implementation. See my comment

src/qfa/domain/models.py
olafdegraaff: I think it would be possible, but then `metadata` would have a pydantic model as type instead of a dict, that means that we (or claude) need to change in many places in the code some syntax, wondering if it improves the code. I can also just get rid of FeedbackRecordMetadata since it was solely for documentation. What do you think?

olafdegraaff: I got rid of it.

scripts/generate_corpus.py
```diff
@@ -775,7 +779,7 @@ def _sample_metadata(
         "source": rng.choices(sources, weights=source_weights, k=1)[0],
         "year": START.year,
         "sensitive": sensitive,
-        "codes": leaf.code_id,
+        METADATA_CODE_FIELD: leaf.code_id,
```
mariushelf, line 782: this shouldn't be configuration. We specify it once. No need to have it dynamic.

src/qfa/domain/models.py
mariushelf: I agree with @teeuwski . I suggest a `FeedbackRecordMetadata` field with some defined, common fields (such as date and code assignments, which also should be a pydantic model).

I was about to suggest a free-form "other" field, but I think that's YAGNI. So let's fully specify any metadata that we expect, and use in at least one endpoint.

Claude can do the heavy lifting for us.

src/qfa/services/prompts.py
```diff
@@ -206,3 +185,35 @@ def build_analyze_judge_system_message(
...
         analysis=analysis,
     )
+
+
+def build_feedback_record_envelope(
+    feedback_record: FeedbackRecordModel,
+) -> str:
+    """Build a single <feedback_record> envelope for a record."""
+    rec_id_attr = _xml_quoteattr(feedback_record.id)
+    rec_text = escape_for_tag_envelope(feedback_record.content)
```
mariushelf, line 195: nit: use `rec_content` in accordance to UL.

src/qfa/settings.py
```diff
@@ -25,6 +25,12 @@
 default and the library default cannot silently drift apart.
 """
 
+# Single source of truth for metadata field names consumed by the pipeline.
+# AnalyzeSettings defaults and the corpus generator both reference these
+# so they cannot silently drift apart.
+METADATA_DATE_FIELD = "created"
+METADATA_CODE_FIELD = "codes"
```
mariushelf, line 32: `settings.py` is for user (or rather, operator) settings.

These constants should go into a `constants.py`.

However, I think we should use a model with attributes instead of a dict, which makes these unnecessary.

mariushelf (changes requested):

mariushelf:
Superseded by #215. Closing.

PR #187, opened 2026-06-17
fix(prompts): unify feedback record prompt shape with XML envelopes

### LINKED ISSUE #131: Unify multi-document feedback-record envelope across all use cases
## Background

Each use case currently invents its own way to lay out feedback records inside the LLM user message. The XML-envelope pattern landed in #126 for `analyze` is the most defensible of the bunch; the others range from "fine, just inconsistent" to "leaks Python internals into the prompt."

This issue tracks a follow-up to unify them around the same building block.

## Current state

| Use case | Multi-document input shape | Escaping | Injection resistance |
|---|---|---|---|
| `analyze` (landed in #126) | XML envelope: `<feedback_record id="…"><text>…</text><metadata>…</metadata></feedback_record>` inside `<feedback_records>…</feedback_records>` | Yes (`escape_for_tag_envelope`) | High |
| `summarize` (per-record) | `str(request.feedback_records)` — Python `repr` of the tuple | No | **None — leaks Python internals** |
| `summarize_aggregate` | `"\n\n".join(f"{idx}. {record.text}" …)` | No | Low |
| `assign_codes` | `"Feedback:\n---\n{feedback_text}\n---\n…"` separator blocks | No | Low |
| `detect_sensitive` | `str(request.feedback_records)` — Python `repr` again | No | **None** |

Five conceptually-identical jobs ("here are feedback records, do something with them"), five different prompt shapes. The two `str(tuple)` paths are the most worrying — an attacker can craft feedback that looks like a legitimate Python `repr` to confuse the model about what counts as a real record.

## Why XML envelopes (not JSON) for the unified shape

This was discussed during the #126 review. Summary of the reasoning:

| Dimension | XML envelope | JSON-in-prompt |
|---|---|---|
| Provider recommendation | Anthropic explicitly recommends XML tags for Claude. OpenAI is provider-neutral but their structured-input cookbooks all use clear delimiters. | No provider explicitly recommends JSON-formatted *input* prompts. |
| Escape semantics | 3 well-defined replacements (`& < >` → entities). Survives copy-paste, logging, display. | Strings need `\"` `\\` `\n` escaping; nested objects compound the problem. Models trained on web text find JSON-escaped text more "code-like" and sometimes hallucinate structure. |
| Role-naming | Tag names label the role of the content (`<analyst_instruction>` vs `<feedback_record>`). That's exactly the trusted/untrusted boundary prompt-injection attacks try to blur. | JSON keys also name fields, but the structural boundary `{ … }` is visually closeable by adversarial content. |

**Critical distinction: input format ≠ output format.** XML for prompt *inputs* (mixed-trust content the LLM is reading); JSON / Pydantic structured outputs for *responses* (parsed by `litellm` + `response_model=…`). The codebase already does this correctly for `analyze` — the proposal extends the input side to the other use cases.

## Proposed direction (deliberately not a spec yet)

1. **Extract a reusable building block** — `build_feedback_record_envelope(record)` and `build_feedback_records_envelope(records)` in `qfa.services.prompts` (or a new `qfa.services.envelopes` module if `prompts.py` grows past the analyse use case).
2. **Migrate all five use cases to it**, keeping each one's task-specific *wrapper* (analyst question, coding context, sensitivity guidance) but unifying the *record* shape.
3. **Drop the `str(tuple)` calls** in `summarize` and `detect_sensitive` — they're the latent injection vector, highest-priority part of the migration.
4. **Route every record id, text, metadata key, metadata value through `escape_for_tag_envelope`** — the `analyze` path already does this; the others currently don't.

## Tradeoffs / risks to surface during design

- **Test churn**: each use case has e2e tests asserting specific LLM-output shapes. Changing prompt shape can shift LLM behaviour at the margin; expect golden-test updates and one or two judgement calls about what counts as "the same" output.
- **Per-record use cases (single-record summarize)**: a full `<feedback_records>` envelope wrapping one `<feedback_record>` is slightly verbose, but the consistency win outweighs the cost.
- **Backwards compatibility**: if any analyst is running these prompts manually with a specific format expectation, unification changes that — worth a quick check with stakeholders before merging.

## Scope (rough)

Roughly one-day effort for a focused PR. Spec / detailed plan to be written when this issue is picked up.

olafdegraaff:
Closes #131 

## Summary

- Extract `build_feedback_record_envelope` and `build_feedback_records_envelope` as reusable building blocks in `qfa.services.prompts`, with `include_metadata` and `include_id` keyword-only flags
- Migrate all five use cases to the unified XML envelope shape:
  - `summarize` — replaces `str(request.feedback_record)` (Python repr, no escaping)
  - `summarize_bulk` — replaces numbered `"\n\n".join(...)` list
  - `detect_sensitive` — replaces `str(request.feedback_record)` (Python repr, no escaping)
  - `assign_codes` — replaces `Feedback:\n---\n{text}\n---` separator blocks in `coding_classifier.py`
  - `analyze` — already on XML envelopes; dead record-building loop removed
- All untrusted strings (record id, text, metadata keys/values) now route through `escape_for_tag_envelope`
- Made the LLM as a judge also rightly implemented for the summarize endpoint. (resolving the TODO)

## Note 
The feedback follows the XML envelop format, but not yet the full LLM message which the LLM receives, e.g.:
for assign code it is:

`  <feedback_record>
    <text>The water distribution was well organized but we had to wait three hours.</text>
  </feedback_record>
Current level:
Code level 1

Options:
0: Service Delivery
1: Health
2: Protection`

do we also want to include that in the #131 ?


src/qfa/services/prompts.py
```diff
@@ -206,3 +184,48 @@ def build_analyze_judge_system_message(
...
+
+
+def build_feedback_record_envelope(
+    feedback_record: FeedbackRecordModel,
+    *,
+    include_metadata: bool = True,
+    include_id: bool = True,
+) -> str:
+    """Build a single <feedback_record> envelope for a record."""
+    rec_text = escape_for_tag_envelope(feedback_record.content)
```
mariushelf: nit: `rec_content` instead of `rec_text`

mariushelf (dismissed):
Approved, with one optional nit to address.

mariushelf (approved):
LGTM!

PR #195, opened 2026-06-19
feat(ci): record deployed versions (tags + GitHub status + run summary) and document checking them

### LINKED ISSUE #197: Release log and deployment status
App and infra are deployed separately from each other and for each environment. 

Currently it is quite hard to see which version of app and infra has been deployed where. 
There is the health endpoint, but if the app is down, that is not reachable. 

Document or implement an easy way of seeing the currently released versions of both app and infra for each environment. 

This is crucial for debugging and monitoring.

mariushelf:
## What

Improves deploy observability and documents how to find what's deployed where.

### Records the deployed version in three places (every deploy)
- **Azure App Service resource tags** — `deployed_version`, `deployed_digest`, `deployed_at`, `deployed_by`. Tags are control-plane metadata, so writing them does **not** restart the app, and `az webapp show --query tags` answers "what's deployed?" even when the container is down. Terraform now ignores these tag keys (`infra/app_service.tf`) so applies won't wipe them.
- **GitHub deployment status description** — `vX.Y.Z @ sha256:…`, posted from a post-deploy `annotate` job so it's the *latest* status and shows on the Environments page (deployments run from `main`, so the bare `ref` was uninformative).
- **Run summary** (`$GITHUB_STEP_SUMMARY`) — a deploy table on the run page, so the Actions list is scannable without opening logs.

Applied in the shared `_deploy-release.yaml` (covers Release → dev, all promotes, auto-staging) and in `build-from-commit.yaml` (so ephemeral dev deploys keep dev's tags truthful). Callers gain `deployments: write` so the reusable workflow can post statuses.

### Tooling + docs
- **`scripts/show_deployed_versions.sh`** — prints app version + infra commit per environment from the GitHub trail (needs `gh`+`jq`, no Azure).
- **`docs/operations/check-deployed-versions.md`** — how to find which app version and infra commit are live per environment, including when the app is down. Wired into the operations index and cross-linked from Release flow.

## Not included (discussed, deferred)
- Stopping Terraform `plan` runs from creating deployment records (needs a repo-level read-only plan identity).
- Separate `infra-*` environments to fully distinguish infra from app deploys.

## Verification
- `make docs` builds clean (no broken refs).
- All workflow YAML parses; yamllint (relaxed) passes via pre-commit.
- `bash -n` clean on the new script.

Closes #197


.github/workflows/_deploy-release.yaml
```diff
@@ -77,3 +79,78 @@ jobs:
...
+        run: |
+          set -euo pipefail
+          DEPLOYMENT_ID=$(gh api \
+            "repos/$GITHUB_REPOSITORY/deployments?environment=$ENVIRONMENT&sha=$GITHUB_SHA&per_page=1" \
+            --jq '.[0].id // empty')
+          if [ -z "$DEPLOYMENT_ID" ]; then
+            echo "::warning::No GitHub deployment found for $ENVIRONMENT @ $GITHUB_SHA; skipping version annotation."
+            exit 0
+          fi
+          gh api "repos/$GITHUB_REPOSITORY/deployments/$DEPLOYMENT_ID/statuses" \
```
mariushelf, line 160: **Correctness (worth fixing before merge):** this status is posted with only `state`/`environment`/`description` — no `log_url`/`target_url`. Two consequences:

1. It becomes the **newest** status on the deployment (that's the whole point of `annotate`). GitHub's own auto-created `success` status *does* carry `target_url`/`log_url` pointing at the run — I confirmed this on a live staging deployment. But both `scripts/show_deployed_versions.sh:41` and `docs/operations/check-deployed-versions.md:176` read the newest status's `target_url` (`statuses?per_page=1` → `.[0].target_url`) to recover the originating workflow name. After this PR that field is empty for every app deploy, so the workflow label resolves to `?`/null — the documented sample output (`[Release]`, `[Auto-deploy…]`) can no longer be produced, and the doc's stated "reliable disambiguator is the originating workflow name … each deployment has a status whose `target_url` points at the workflow run" is falsified for app deploys by this very change. (Infra detection still works, since Terraform deployments aren't annotated.)
2. On the **Environments page**, GitHub shows the latest status. Without `log_url`/`environment_url`, the "View deployment" run link that the auto-success status carried disappears.

**Both fixed with one line** — give the annotate status a `log_url` so it keeps pointing at the run:
```bash
gh api "repos/$GITHUB_REPOSITORY/deployments/$DEPLOYMENT_ID/statuses" \
  -f state=success \
  -f environment="$ENVIRONMENT" \
  -f description="$TAG @ $DIGEST" \
  -f log_url="$GITHUB_SERVER_URL/$GITHUB_REPOSITORY/actions/runs/$GITHUB_RUN_ID" \
  --silent
```
This keeps `target_url` populated on the latest status (so the script + doc loop keep resolving the workflow name) and restores the run link on the Environments page. Alternatively/additionally, make the script and doc snippet select the *first status with a non-empty `target_url`* (mirroring how line 56-57 already handles `description`) — but the `log_url` fix is preferable because it keeps the doc's stated invariant true and fixes the Environments-page link too.

scripts/show_deployed_versions.sh
```diff
@@ -0,0 +1,62 @@
...
+
+for env in "${ENVIRONMENTS[@]}"; do
+  printf '== %s ==\n' "$env"
+  app_done=false
+  infra_done=false
+
+  for id in $(gh api "repos/$REPO/deployments?environment=$env&per_page=20" --jq '.[].id'); do
+    $app_done && $infra_done && break
+
+    target_url=$(gh api "repos/$REPO/deployments/$id/statuses?per_page=1" --jq '.[0].target_url // ""')
```
mariushelf: **Correctness (see the headline on `_deploy-release.yaml`):** `.[0].target_url` reads the *newest* status, which after this PR is the `annotate` status that carries no `target_url` — so `run_id` comes back empty, the `actions/runs/` lookup fails, and every app row's `[workflow]` label degrades to `?`. Note the asymmetry with line 56-57, which already robustly does `first(.[] | select(.description != "") | .description)` for the version. Apply the same shape here:
```bash
target_url=$(gh api "repos/$REPO/deployments/$id/statuses" \
  --jq 'first(.[] | select(.target_url != "") | .target_url) // ""')
```
(Preferred: add `log_url` to the annotate status as described on `_deploy-release.yaml` — then this line works unchanged *and* the Environments page keeps its run link.)

docs/operations/check-deployed-versions.md
```diff
@@ -0,0 +1,188 @@
...
+(see [Quick answer](#quick-answer-the-helper-script) above). The loop it wraps,
+for reference — it labels every recent deployment per environment with the
+workflow that produced it, separating infra applies from app deploys and
+exposing the `ref`/`sha` and timestamp in one pass:
+
+```bash
+for env in dev staging prd; do
+  echo "===== $env ====="
+  for id in $(gh api "repos/{owner}/{repo}/deployments?environment=$env&per_page=5" --jq '.[].id'); do
+    url=$(gh api "repos/{owner}/{repo}/deployments/$id/statuses?per_page=1" --jq '.[0].target_url')
```
mariushelf: **Doc correctness:** same `.[0].target_url` pattern as the script — after this PR the newest status is the description-only `annotate` status with no `target_url`, so this reference loop prints `[]`/empty for the workflow name on app deploys. Whichever fix you take for the script, mirror it here, and double-check the sample output at lines 19-22 (`[Release]`, `[Auto-deploy published release to staging]`) still reflects reality — as written they can't be reproduced until the `target_url`/`log_url` gap is closed.

.github/workflows/_deploy-release.yaml
```diff
@@ -77,3 +79,78 @@ jobs:
             --resource-group "${{ vars.AZ_RESOURCE_GROUP }}" \
             --container-image-name "${IMAGE}@${DIGEST}" \
             --container-registry-url "https://${{ vars.AZ_ACR_NAME }}.azurecr.io"
+
+      - name: Stamp version onto App Service tags and run summary
```
mariushelf, line 84: **Consider (robustness):** this stamp step runs *after* the container swap in the same `deploy` job under `set -euo pipefail`. If `az webapp update` (tags) or the `$GITHUB_STEP_SUMMARY` write fails, the step fails → the whole `deploy` job fails → GitHub marks the deployment **failed** and the `annotate` job is skipped — even though the app was already deployed successfully. An observability annotation shouldn't be able to turn a real successful deploy red (and could confuse alerting/rollback). Suggest making the stamp best-effort, e.g. wrap the `az webapp update` in `|| echo "::warning::failed to stamp deployed-version tags"`, or split it into its own `needs: deploy` job that can fail independently. The tags are advisory metadata, so a warning is the right severity. (Same applies to the twin step in `build-from-commit.yaml`.)

mariushelf (commented):
## Review: record deployed versions (approach + correctness)

**Approach — good idea, follows best practice.** Recording the deployed version in three durable places is sound defense-in-depth, and each place is chosen for the right reason:
- **Azure resource tags** are control-plane metadata, so they survive a container that won't start and don't trigger a restart (unlike app settings). Correct call.
- **GitHub deployment status** is the natural home for "what's live", and posting it from a separate post-deploy `annotate` job so it becomes the *latest* status is exactly right given GitHub writes its own final status when the deploy job ends. The reasoning in the job comment is accurate.
- **Run summary** for scannability. Fine.
- Terraform `ignore_changes` to stop applies wiping the tags is the correct mechanism, and the new steps correctly pass user-controlled values (`TAG`, `DIGEST`, `GITHUB_ACTOR`) through `env:` rather than inlining them into the shell — no script-injection surface. 👍

So the design is worth shipping. The findings below are about the implementation and docs, ordered by leverage.

**Headline:** the `annotate` status is posted without a `log_url`/`target_url`, which (a) breaks the workflow-name disambiguation that both the helper script and the how-to doc rely on, and (b) drops the "View deployment" link from the latest status on the Environments page. Details inline on `_deploy-release.yaml`. This one is worth fixing before merge because it partially undermines the very tooling this PR ships. The rest are Consider/Optional.

Verified: rebased cleanly onto current `origin/main`; all doc cross-refs resolve; `bash -n` clean; YAML parses. I confirmed against the live repo API that GitHub's auto-created `in_progress`/`success` statuses carry `target_url`+`log_url`, whereas the `annotate` POST sets only `state`/`environment`/`description`.

.github/workflows/build-from-commit.yaml
```diff
@@ -96,3 +96,37 @@ jobs:
             --resource-group "${{ vars.AZ_RESOURCE_GROUP }}" \
             --container-image-name "${IMAGE}@${DIGEST}" \
             --container-registry-url "https://${{ vars.AZ_ACR_NAME }}.azurecr.io"
+
+      - name: Stamp ephemeral version onto App Service tags and run summary
```
mariushelf, line 101: **Consider (DRY):** this ~25-line stamp block (tag update + summary table) is a near-verbatim copy of the one in `_deploy-release.yaml:83` — the wording already differs slightly ("ephemeral"), which is exactly how these drift. Reusable-workflow *steps* can't be shared directly, but a **composite action** can: `.github/actions/stamp-deployed-version/action.yml` taking `app-name`/`resource-group`/`tag`/`digest`/`environment` inputs, called from both places. Optional at two copies, but it's the canonical move and keeps the tag schema in one spot.

scripts/show_deployed_versions.sh
```diff
@@ -0,0 +1,62 @@
...
+run_id_from_url() {
+  printf '%s' "$1" | grep -oE 'runs/[0-9]+' | grep -oE '[0-9]+' | head -n1
+}
+
+for env in "${ENVIRONMENTS[@]}"; do
+  printf '== %s ==\n' "$env"
+  app_done=false
+  infra_done=false
+
+  for id in $(gh api "repos/$REPO/deployments?environment=$env&per_page=20" --jq '.[].id'); do
```
mariushelf, line 38: **Optional (silent truncation):** this reads only the 20 most recent deployments per environment. Since Terraform `plan` runs also create deployment records (acknowledged in the PR body and in the how-to), a burst of plans + app deploys can push the last real `workflow_dispatch` **apply** past the 20-record window — and then the `infra:` line silently never prints for that env, which reads as "no infra info" rather than "looked but didn't find one." Either paginate until an apply is found (bounded), or emit an explicit `infra: <none found in last 20 deployments>` line so the absence is visible instead of silent.

infra/app_service.tf
```diff
@@ -84,7 +84,14 @@ resource "azurerm_linux_web_app" "backend" {
...
   lifecycle {
     # The container image tag is updated by the CI/CD pipeline, not Terraform
-    ignore_changes = [site_config[0].application_stack]
+    ignore_changes = [
+      site_config[0].application_stack,
+      # The deploy pipeline writes deployed_version / deployed_digest /
+      # deployed_at / deployed_by tags on every deploy (see
+      # .github/workflows/_deploy-release.yaml). Terraform must not manage or
+      # revert them, or each apply would wipe the deployed-version record.
+      tags,
```
mariushelf: **Optional (latent footgun):** `ignore_changes = [tags]` ignores the *entire* tags map, but the comment above says "these tag keys" (the four `deployed_*` ones). It's harmless today because this resource declares no `tags = {}` block — but the moment someone adds Terraform-managed tags (cost-center, `managed-by`, env), Terraform will silently stop reconciling them and the drift becomes invisible. Terraform supports per-key ignore for maps, so scope it to intent:
```hcl
ignore_changes = [
  site_config[0].application_stack,
  tags["deployed_version"],
  tags["deployed_digest"],
  tags["deployed_at"],
  tags["deployed_by"],
]
```

mariushelf:
Addressed all seven review findings in 79f5ed2:

| # | Finding | Fix |
|---|---|---|
| 1 | annotate status drops `target_url`/run link (headline) | annotate POST now sets `-f log_url=…/actions/runs/$GITHUB_RUN_ID`; GitHub mirrors it into `target_url`, restoring the Environments-page link and the tooling lookup |
| 2 | script reads newest status's `target_url` | now selects the first status *with* a `target_url` (`first(.[] | select(.target_url != "") | .target_url)`) |
| 3 | doc reference loop had the same pattern | mirrored the fix; noted the annotate status carries the link via `log_url` |
| 4 | stamp failure could fail a real deploy | stamp step is `continue-on-error: true` and the `az` update is `|| ::warning::` best-effort |
| 5 | ~25-line stamp block duplicated across two workflows | extracted to composite action `.github/actions/stamp-deployed-version`, called from both (each job gains a sparse checkout so the local action is on disk) |
| 6 | `per_page=20` could silently hide an infra apply | prints an explicit `<none found in last 20 deployment records>` line per category when the window has none |
| 7 | `ignore_changes = [tags]` too broad | scoped to `tags["deployed_version"]` / `_digest` / `_at` / `_by` |

Verified: `make docs` builds clean under `-W`; `bash -n` clean; all workflow + composite-action YAML parses; yamllint passes via pre-commit. `terraform fmt/validate` not run locally (terraform not installed in this environment) — CI will cover it.

mariushelf (commented):
_(Duplicate — this review posted twice due to a transient API hiccup. See the identical re-review just below; its inline FYI is the canonical one.)_

.github/workflows/_deploy-release.yaml
```diff
@@ -77,3 +79,64 @@ jobs:
             --resource-group "${{ vars.AZ_RESOURCE_GROUP }}" \
             --container-image-name "${IMAGE}@${DIGEST}" \
             --container-registry-url "https://${{ vars.AZ_ACR_NAME }}.azurecr.io"
+
+      # Needed only so the local composite action below is on disk (this job
+      # otherwise works purely via the gh/az APIs and needs no source).
+      - name: Check out for the stamp action
```
mariushelf: **FYI (non-blocking, a consequence of going local-composite-action):** switching from an inline `run:` block to `uses: ./.github/actions/…` means this job now checks out source, and with no `ref:` that's `github.sha` — i.e. the **deployed ref**. For `promote-to-*` (workflow_dispatch from `main`) that's `main`, which always has the action. But for the `release`/publish-triggered paths, `github.sha` is the *release tag's commit*; re-deploying or re-publishing a release cut **before this PR merged** would check out a commit without `.github/actions/stamp-deployed-version/`, so the stamp step can't resolve.

That degrades gracefully — `continue-on-error: true` means the deploy still succeeds and the `annotate` job (no checkout) still records the version on the GitHub deployment; only the Azure tags + run-summary table are skipped for those legacy redeploys. So nothing to fix. If you ever wanted to fully decouple the tooling version from the deployed artifact, pin the checkout to a stable ref (e.g. `ref: ${{ github.event.repository.default_branch }}`) — but given it's advisory metadata and self-heals for all releases cut from here on, leaving it is reasonable.

mariushelf (commented):
## Re-review of the fix commit (79f5ed2)

Walked the fix with fresh eyes against all five axes. **All seven prior findings are correctly resolved** and I verified the result builds:

- **Headline (annotate `target_url`)** — the `annotate` status now sets `log_url`, and independently the script/doc select the *first status that has a `target_url`* rather than the newest. That's belt-and-suspenders: even if `log_url`→`target_url` mirroring ever didn't happen, the readers fall back to GitHub's own auto-`success` status (which always carries `target_url`), so the workflow-name lookup is correct either way. Good.
- **Best-effort stamping** — `continue-on-error: true` keeps the *job* green, and the internal `az … || ::warning::` ensures the run-summary table still writes even if only the tag write fails. Two distinct failure surfaces, both covered.
- **DRY** — the tag schema (`deployed_version`/`_digest`/`_at`/`_by`) now lives in exactly one place; a future schema change is a one-file edit. Justified extraction.
- **Silent cap**, **Terraform key scoping**, **doc mirror** — all correct.

**Verification:** `make docs` builds clean under `-W` (warnings-as-errors); all workflow + composite-action YAML parses; yamllint passes via pre-commit; `bash -n` clean. `terraform fmt/validate` not run (terraform absent locally) — CI covers it.

One **FYI** inline (non-blocking) on the new checkout step. Net verdict: **approve** — this is a clear improvement over the first pass.

.github/workflows/_deploy-release.yaml
mariushelf: ✅ Fixed in 79f5ed2 — the annotate POST now sets `-f log_url="$GITHUB_SERVER_URL/$GITHUB_REPOSITORY/actions/runs/$GITHUB_RUN_ID"`, so the latest status keeps the run link (Environments-page button) and a populated `target_url`.

scripts/show_deployed_versions.sh
mariushelf: ✅ Fixed in 79f5ed2 — now selects the first status that has a `target_url` (`first(.[] | select(.target_url != "") | .target_url)`) instead of blindly the newest, matching the `description` handling. Robust even if the annotate status ever lacked one (falls back to GitHub's auto-`success` status).

docs/operations/check-deployed-versions.md
mariushelf: ✅ Fixed in 79f5ed2 — mirrored the first-non-empty-`target_url` selection in the reference loop, and noted that the annotate status carries the link via `log_url`. Sample output at the top is now reproducible.

.github/workflows/_deploy-release.yaml
mariushelf: ✅ Fixed in 79f5ed2 — stamping is now `continue-on-error: true` and the `az webapp update` inside the shared action is `|| echo "::warning::…"`, so a failure to write advisory tags can never turn a successful deploy red.

.github/workflows/build-from-commit.yaml
mariushelf: ✅ Done in 79f5ed2 — extracted to a composite action `.github/actions/stamp-deployed-version`, called from both deploy paths, so the tag schema lives in one place. Each job gains a small `sparse-checkout` so the local action is on disk; see the FYI on the checkout step re: the deployed-ref coupling (non-blocking, self-heals for releases cut from here on).

scripts/show_deployed_versions.sh
mariushelf: ✅ Fixed in 79f5ed2 — the script now prints an explicit `<none found in last 20 deployment records>` line per category when the window contains none, so absence is visible rather than silent.

infra/app_service.tf
mariushelf: ✅ Fixed in 79f5ed2 — scoped to `tags["deployed_version"]` / `tags["deployed_digest"]` / `tags["deployed_at"]` / `tags["deployed_by"]`, so any Terraform-managed tags added later are still reconciled.

.github/workflows/build-from-commit.yaml
mariushelf: ↩️ Reverted in e13c90e, per author call. Agreed on reflection: for two stable call sites the composite action was net *more* complex — a line increase across three files, extra concepts (composite action, sparse-checkout, an `ephemeral` flag), and a real behavior change (the checkout-at-deployed-ref coupling in the FYI below). The two ~25-line stamp steps are back inline and duplicated; the best-effort behavior (`continue-on-error` + warning) is preserved inline in both. The only cost is the 4-key tag schema now living in two places, which is cheap to keep in sync.

.github/workflows/_deploy-release.yaml
mariushelf: Moot as of e13c90e — the checkout step is gone (the stamp is inline again), so there is no longer any deployed-ref coupling.

PR #198, opened 2026-06-19
docs(operations): add how-to runbook for forcing Key Vault reference refresh

mariushelf:
## What

Adds a new operations runbook page, **Operational How-Tos** (`docs/operations/how-to.md`), with its first entry: how to force App Service to re-read changed Key Vault values.

## Why

App Service caches Key Vault references, so a rotated secret doesn't take effect until the worker recycles. There was no documented, copy-pasteable procedure for forcing an immediate refresh — only a one-line aside in `auth-management.md`.

## Contents

- **Environment naming table** — clarifies the asymmetric `-n` / `-g` naming so commands aren't accidentally run against the wrong environment:
  | Env | App Service (`-n`) | Resource group (`-g`) | Key Vault |
  |---|---|---|---|
  | dev | `qfa-dev-backend` | `qualitative-feedback-analysis-xomnia` | `qfa-dev-keyvault` |
  | staging | `qfa-staging-backend` | `qualitative-feedback-analysis-staging` | `qfa-staging-keyvault` |
  | prd | `qfa-prd-backend` | `qualitative-feedback-analysis-production` | `qfa-prd-keyvault` |
- **Force refresh** — the `az webapp config appsettings set … --settings KV_REFRESH_TOUCH=$(date +%s)` recipe, plus restart-only and Portal alternatives.
- **CLI token-staleness tip** for the `AuthorizationFailed` case after a recent role grant / PIM activation.
- **Versionless vs. pinned references** note.

## Wiring

- Added to the operations index table and `{toctree}`.
- Cross-linked the existing note in `auth-management.md` to the new section.

## Verification

`make docs` builds cleanly — the new `operations/how-to` page is picked up with no toctree or broken-link warnings.

### DIFF
--- a/docs/operations/auth-management.md
+++ b/docs/operations/auth-management.md
@@ -92,7 +92,7 @@ Key rotation steps:
 3. Wait until you've verified the tenant is using the new key (look for the new `key_id` in logs).
 4. `--replace <tenant>` with their now-confirmed new key, or `--remove <tenant>` plus `--add` if you want a clean cut.
 
-App Service picks up Key Vault changes within a few minutes. To force an immediate refresh, see [Operational how-tos § Force refresh of changed Key Vault values](how-to.md#force-refresh-of-changed-key-vault-values).
+App Service can take up to ~24 hours to pick up a rotated Key Vault value on its own — it caches references and refetches them only about once a day. To force an immediate refresh, see [Operational how-tos § Force refresh of changed Key Vault values](how-to.md#force-refresh-of-changed-key-vault-values).
 
 ## Local dev
 
--- a/docs/operations/auth-management.md
+++ b/docs/operations/auth-management.md
@@ -92,7 +92,7 @@ Key rotation steps:
 3. Wait until you've verified the tenant is using the new key (look for the new `key_id` in logs).
 4. `--replace <tenant>` with their now-confirmed new key, or `--remove <tenant>` plus `--add` if you want a clean cut.
 
-App Service picks up Key Vault changes within a few minutes. To force an immediate refresh, restart the App Service from the Azure portal.
+App Service picks up Key Vault changes within a few minutes. To force an immediate refresh, see [Operational how-tos § Force refresh of changed Key Vault values](how-to.md#force-refresh-of-changed-key-vault-values).
 
 ## Local dev
 
--- a/docs/operations/how-to.md
+++ b/docs/operations/how-to.md
@@ -1,0 +1,67 @@
+# Operational How-Tos
+
+Short, copy-pasteable runbooks for recurring operational tasks. Each entry is
+self-contained; pick the one you need.
+
+## Environment naming
+
+The resource group and App Service names are **not** symmetric across
+environments, so double-check both before running a command against production.
+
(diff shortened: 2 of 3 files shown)


docs/operations/auth-management.md
```diff
@@ -92,7 +92,7 @@ Key rotation steps:
 3. Wait until you've verified the tenant is using the new key (look for the new `key_id` in logs).
 4. `--replace <tenant>` with their now-confirmed new key, or `--remove <tenant>` plus `--add` if you want a clean cut.
 
-App Service picks up Key Vault changes within a few minutes. To force an immediate refresh, restart the App Service from the Azure portal.
+App Service picks up Key Vault changes within a few minutes. To force an immediate refresh, see [Operational how-tos § Force refresh of changed Key Vault values](how-to.md#force-refresh-of-changed-key-vault-values).
```
mariushelf: is this claim true (picks up within minutes)???

From my experience it is not.

docs/operations/how-to.md
```diff
@@ -0,0 +1,67 @@
...
+
+`KV_REFRESH_TOUCH` is an arbitrary, unused setting; the `$(date +%s)` value just
+guarantees it changes each run so the save always recycles the app. The app
+re-resolves **all** Key Vault references on the restart that follows.
+
+Equivalent alternatives:
+
+- **Restart only** (leaves no throwaway setting behind):
+  ```bash
+  az webapp restart -n qfa-prd-backend -g qualitative-feedback-analysis-production
```
mariushelf: I think this will NOT refresh from the vault

docs/operations/auth-management.md
mariushelf: You're right — the claim was wrong. Microsoft documents the auto-refresh interval as **~24 hours**, not minutes: App Service *"caches the values of the Key Vault references and refetches them every 24 hours"*, and a rotated (versionless) secret *"is automatically updated ... within 24 hours"* ([Understand rotation](https://learn.microsoft.com/en-us/azure/app-service/app-service-key-vault-references#understand-rotation)). There's no evidence the interval was ever "a few minutes."

Fixed in 65f5430 — this line now reads "can take up to ~24 hours ... refetches them only about once a day", and the same wrong figure in `how-to.md` is corrected too.

docs/operations/how-to.md
mariushelf: Confirmed — a bare `az webapp restart` is **not** documented to re-read Key Vault references, and community reports (several Microsoft Q&A threads, an Azure `azure-functions-host` issue) say it often keeps serving the cached values. The docs only credit a *configuration change*: *"Any configuration change to the app causes an app restart and an immediate refetch of all referenced secrets."*

So the `appsettings set KV_REFRESH_TOUCH=…` recipe above is correct (it's a config change), but the "Restart only" alternative was not equivalent. Fixed in 65f5430:
- Removed the "Restart only" alternative.
- Added the documented **refresh API** (`.../config/configreferences/appsettings/refresh`) as the clean no-throwaway-setting option.
- Added an explicit "don't rely on a bare restart" warning.
- Reworded the surrounding text to attribute the refetch to the config change, not the restart.

PR #208, opened 2026-06-24
feat(postgres): guard database resources against accidental destruction

### LINKED ISSUE #180: Data-bearing Postgres resources lack prevent_destroy — verify exposure to accidental recreate
Context: Azure Postgres flexible server defined in `postgres.tf`. The server itself has `lifecycle { prevent_destroy = true }`; the resources around it do not. **The scenarios below are inferred — please verify they apply before acting (see disclaimer).**

## Problem (to verify)

`prevent_destroy` guards the Postgres **server** only. Not guarded in `postgres.tf`: the application **database** (holds the data), the **delegated subnet**, the **private DNS zone**, and the **DNS-zone vnet link**. Inferred risks:

- Changing the database name / collation / charset forces a **database** replace → drops and recreates it empty while the server survives (looks safe, data gone).
- A delegated-subnet prefix change forces the server to detach/recreate.
- Recreating the private DNS zone (e.g. a wrong-workspace apply) silently breaks DB name resolution at runtime, with no Terraform error.
- `storage_mb` can't be lowered, so a downward change wedges future applies.

## Disclaimer

These come from reading the config + general Azure provider behavior, not from a reproduction. **Investigate whether each scenario actually forces a destroy/replace here, and whether it's relevant given the team's apply workflow** — some may never occur in practice (e.g. these attributes are never changed).

## Suggested fix (if confirmed)

- Add `lifecycle { prevent_destroy = true }` to the database, delegated subnet, private DNS zone, and vnet link.
- Add a `storage_mb` precondition rejecting values below the current size.
- Treat db name / collation / charset as immutable (`ignore_changes`).

olafdegraaff:
Closes #180 (see comments at the ticket)

## Summary

- **`azurerm_private_dns_zone`**: add `prevent_destroy = true` — deleting the zone silently breaks DB hostname resolution at runtime with no Terraform error (Azure allows the deletion even while the server references it)
- **`azurerm_postgresql_flexible_server_database`**: add `prevent_destroy = true` and `ignore_changes = [name, collation, charset]` — the database holds the actual data; these three attributes are immutable in Azure (any change forces a destroy+create), so ignoring them prevents an accidental config change from dropping and recreating the database
- **`azurerm_postgresql_flexible_server`**: add a comment on `storage_mb` clarifying that Azure rejects any downward change, making a reduction in config permanently wedge future applies

### DIFF
--- a/infra/postgres.tf
+++ b/infra/postgres.tf
@@ -21,6 +21,10 @@ resource "azurerm_subnet" "qfa_db_snet" {
 resource "azurerm_private_dns_zone" "postgres" {
   resource_group_name = data.azurerm_resource_group.main.name
   name                = "qfa-${local.env}.postgres.database.azure.com"
+
+  lifecycle {
+    prevent_destroy = true
+  }
 }
 
 resource "azurerm_private_dns_zone_virtual_network_link" "postgres_vnet_link" {
@@ -46,7 +50,7 @@ resource "azurerm_postgresql_flexible_server" "db" {
     password_auth_enabled         = false
     tenant_id                     = var.tenant_id
   }
-
+  # Azure does not allow lowering storage - only increases are accepted.
   storage_mb            = var.postgres_storage_mb
   sku_name              = var.postgres_sku_name
   backup_retention_days = 30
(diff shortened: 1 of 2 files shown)


mariushelf (approved):
LGTM.

PR #211, opened 2026-06-25
feat(framework): support variable-depth coding frameworks in assign-codes

### LINKED ISSUE #160: Decide how to handle `assign-codes` frameworks that don't follow correct shape
Currently, the framework has a check that checks if every leaf node is at 3 deep (codinglevel1, 2, and 3 are present exactly). If this is not followed, a 422 unprocessible entity is thrown.

We should decide if this is the way forward or we want to have other functionalities trigger when we don't have exactly 3.

olafdegraaff:
Closes #160 

As discussed in the meeting this morning (25-06-2026). We should be able to process frameworks where the codelevel sometimes stops at level 1 and 2. Further, we can assume that the coding frameworks cover all kinds of feedback, e.g. each feedback has at least one matching code. This is changed in the SYSTEM_PROMPT of the LLM.

## Summary

- Removed the 3-level depth requirement from `ApiCodingFramework` — frameworks with 1, 2, or any number of levels are now accepted
- Changed `AssignedCodeModel` and `ApiAssignedCode` to flat fields with nullable `coding_level_2_*` and `coding_level_3_*` — the response shape is unchanged for 3-level frameworks; missing levels return `null`
- Replaced the hardcoded 3-level nested loop in `Orchestrator.assign_codes` with a recursive `_traverse_coding_level` helper that treats any node with no children as a leaf, regardless of depth

## Test plan

- [x] `make test` passes (158 API tests, 502 total)
- [x] `make lint` clean
- [x] Call assign-codes with a 1-level framework → `coding_level_2_*` and `coding_level_3_*` are `null`
- [x] Call assign-codes with a 2-level framework → `coding_level_3_*` is `null`
- [x] Call assign-codes with a 3-level framework → all fields populated (existing behaviour unchanged)


teeuwski (approved):
LGTM, really happy about seeing recursion here now. This was bothering me before ;)

PR #215, opened 2026-07-01
Feat/formalise feedback record 2

### LINKED ISSUE #143: Formalise the feedback-record metadata contract

olafdegraaff:
Replacement of PR #186 (thought it was better to start over on a new branch)
Closes #143 

The metadata that is passed to the model is: created, coding_level_1, coding_level_2, coding_level_3.

The metadata is created in the flowchart in espoCRM. For now I think that the given fields cover our current use-cases. In case that other use-cases arise in the future we can add more fields.

## Summary

- **Formalize `FeedbackRecordMetadataModel`** — replaces the previous untyped `dict[str, str | int | float | bool]` metadata field with a typed Pydantic model. Named fields (`created`, `coding_level_1`, `coding_level_2`, `coding_level_3`) match the EspoCRM pipeline convention; `created` defaults to `""` (no date) so existing callers that omit it continue to work.
- **Align settings and prompts** — `OrchestratorSettings.coding_trend_code_fields` now defaults to `["coding_level_1", "coding_level_2", "coding_level_3"]` instead of the old `["codes"]`. The prompt builder switches to `exclude_defaults=True` so empty/unset metadata fields are omitted from the LLM context.
- **Update fixtures and tests** — `fixtures/large_corpus.yaml` updated to use `coding_level_1` instead of `codes`; all test helpers and inline constructions updated to `FeedbackRecordMetadataModel(...)` (type-safe, passes `make ty`).

## Test plan

- [x] `make test` — 504 tests pass, 0 failures
- [x] `make ty` — no type errors
- [x] `make lint` — no lint errors
- [ ] Deploy to dev and verify `/v1/analyze` request with EspoCRM metadata (`created`, `coding_level_1`) returns a coding trend table in the response


src/qfa/services/clustering.py
```diff
@@ -65,7 +65,7 @@ def _sort_by_date(
         return records
 
     def key(record: FeedbackRecordModel) -> tuple[bool, str]:
-        prefix = _iso_date_prefix(record.metadata.get(date_field))
+        prefix = _iso_date_prefix(getattr(record.metadata, date_field, None))
```
mariushelf: since the date field is formalised now, we can simply use `record.metadata.created` now.

src/qfa/api/routes.py
```diff
@@ -185,7 +186,11 @@ async def analyze_bulk(
         )
 
     domain_feedback_records = tuple(
-        FeedbackRecordModel(id=doc.id, content=doc.content, metadata=doc.metadata)
+        FeedbackRecordModel(
+            id=doc.id,
+            content=doc.content,
+            metadata=FeedbackRecordMetadataModel.model_validate(doc.metadata),
```
mariushelf: **Request-schema / domain type mismatch can turn valid requests into HTTP 500.**

The public request schema still declares `metadata: dict[str, str | int | float | bool]` (`src/qfa/api/schemas.py:334`), so a body like `{"metadata": {"created": 2024}}` or `{"metadata": {"coding_level_1": 5}}` passes FastAPI request validation. This line then calls `FeedbackRecordMetadataModel.model_validate(...)`, where `created: str` / `coding_level_1: str | None` reject the int — Pydantic v2 does **not** coerce int→str. That raises a raw `pydantic.ValidationError`, which is *not* a FastAPI `RequestValidationError`, so it bypasses `_handle_validation_error` and falls through to the generic `Exception` handler → **500 instead of 422**.

The same pattern appears in `summarize_bulk`, `summarize`, `assign_codes`, and `detect_sensitive`. Note the schema's own docstring cites #138, where over-strict validation silently broke entire batches — this reintroduces that failure mode in a worse form (500, not 422). Suggest either tightening the request schema so FastAPI produces a proper 422, or sanitising/coercing the incoming dict in the route before `model_validate`.

~Claude

src/qfa/domain/models.py
```diff
@@ -19,6 +19,38 @@
...
+class FeedbackRecordMetadataModel(BaseModel):
+    """Metadata associated with a feedback record.
+
+    Named fields (`created`, `feedback_record_id`, `codes`) match the EspoCRM
+    pipeline convention (see `scripts/espo_crm/`). Additional arbitrary
+    key-value pairs are accepted and round-tripped transparently so that
+    corpus-level fields (e.g. `region`, `year`) continue to work.
+    """
+
+    model_config = ConfigDict(frozen=True)
```
mariushelf: **Extra metadata keys are silently dropped — this contradicts the docstring above.**

The docstring states extra key-value pairs are *"accepted and round-tripped transparently so that corpus-level fields (e.g. `region`, `year`) continue to work."* But `model_config = ConfigDict(frozen=True)` sets no `extra=` option, so Pydantic's default `extra="ignore"` applies: any field outside the four named ones is **discarded** with no error. A client sending `{"region": "North", "coding_level_1": "Water"}` loses `region` silently, and it never reaches the LLM prompt.

This is a behaviour regression from the old arbitrary-dict metadata (the deleted `test_escapes_metadata_keys_and_values` test is direct evidence the capability was removed). Either add `extra="allow"` to genuinely round-trip extra fields as the docstring promises, or update the docstring to state that unknown fields are dropped.

~Claude

src/qfa/domain/models.py
```diff
@@ -19,6 +19,38 @@
 from qfa.domain.sensitivity_types import SensitivityType
 
 
+class FeedbackRecordMetadataModel(BaseModel):
+    """Metadata associated with a feedback record.
+
+    Named fields (`created`, `feedback_record_id`, `codes`) match the EspoCRM
```
mariushelf: **Docstring names fields that don't exist on the model.**

This line lists the named fields as `created`, `feedback_record_id`, `codes` — but the model actually defines `created`, `coding_level_1`, `coding_level_2`, `coding_level_3`. Neither `feedback_record_id` nor `codes` is a field here (and the updated EspoCRM triggers no longer emit `feedback_record_id` in the metadata). Please update the docstring to match the real fields so it doesn't mislead the next reader.

~Claude

src/qfa/settings.py
```diff
@@ -228,7 +228,7 @@ class AnalyzeSettings(BaseSettings):
         description="Metadata key holding the record date for the coding-trend table.",
     )
     coding_trend_code_fields: list[str] = Field(
-        default_factory=lambda: ["codes"],
+        default_factory=lambda: ["coding_level_1", "coding_level_2", "coding_level_3"],
```
mariushelf, line 227: **Docs are now out of date with this default and need updating.**

- `docs/operations/settings-reference.md` (the `ANALYZE_CODING_TREND_CODE_FIELDS` row) still lists the default as `["codes"]` — after this change it's `["coding_level_1", "coding_level_2", "coding_level_3"]`. An operator copying that table would configure the wrong field names and get an empty coding-trend table.
- `docs/ubiquitous_language.md` ("Feedback metadata") still describes metadata as open-ended contextual attributes ("operational and ethical flags", "extra fields"). If the strict typed model stays (see the `extra="allow"` comment on `models.py`), this entry over-promises and needs a caveat; if extra fields are allowed through, it can stay as-is.

~Claude

src/qfa/domain/models.py
mariushelf: Human comment: IMHO we should update and formalise the api model, too, so that the API does not allow extra fields, and correctly documents the expected format.

src/qfa/api/routes.py
mariushelf: human comment: I'd suggest to formalise the API model, too.

mariushelf (changes requested):
The changes LGTM, with some nits, and IMHO a missing part: updating / formalising the Api side of the models (e.g., `ApiFeedbackRecordInput`). There, the metadata can, and IMHO should, also be formalised.

Documentation should be updated.

Now that we have a formalised model, we don't need `get` or `getattr` anymore to access the coding fields.

olafdegraaff:
Also formalized the API model now. Following the ADR-007, i have not reused the domain model but have created a API schema independent of (but in line with) the domain model.

mariushelf (approved):
LGTM now!

PR #221, opened 2026-07-10
feat(metadata): accept deprecated feedback_record_id for EspoCRM compat

mariushelf:
## Summary

Re-admits `feedback_record_id` as a **deprecated, ignored** field on both the API metadata model (`ApiFeedbackRecordMetadata`) and the domain metadata model (`FeedbackRecordMetadataModel`).

### Why
The formalized metadata schema (since ~v2.0.1) locks the request `metadata` object to `{created, coding_level_1, coding_level_2, coding_level_3}` with `extra="forbid"`. Pre-v2.0.1 EspoCRM flowcharts also wrote `feedback_record_id` into metadata (a copy of the record-level `id`). That key is the **only** element of an old EspoCRM payload the current API rejects — so deploying the backend against an un-upgraded EspoCRM would `422` **every** feedback and insight save.

`feedback_record_id` was always redundant: the backend correlates results by the record-level `id` and never read the metadata copy. So the field is accepted purely for backward compatibility and is ignored.

### Design choice
Declared the single legacy key explicitly instead of switching the model to `extra="ignore"`, so the schema still fails fast on genuinely unknown keys (typo-catching preserved). The field is also excluded from LLM prompt rendering (`prompts.py`) so it never reaches the model — it would only duplicate the record `id` already emitted on the `<feedback_record>` tag.

### Effect
The backend can now be deployed **independently** of the EspoCRM flowchart upgrade (and vice versa): old flowcharts (send the key) and new flowcharts (omit it) are both valid.

## Changes
- `src/qfa/api/schemas.py`, `src/qfa/domain/models.py`: add deprecated optional `feedback_record_id` field with rationale in docstrings.
- `src/qfa/services/prompts.py`: exclude `feedback_record_id` from the rendered `<metadata>` block.
- `tests/domain/test_models.py`: legacy field accepted; unknown keys still rejected.
- `tests/api/test_routes.py`: full old-format EspoCRM payload → `200` (not `422`).
- `docs/integrations/espo-crm.md`: note on backend/flowchart deploy independence.

## Verification
- `make test` → 525 passed, 42 deselected
- `make lint` → all checks passed (ruff, ty, import-linter contracts kept)

## Not addressed (out of scope)
Response-side nullable `coding_level_2_id`/`coding_level_3_id` in `assign-codes` (safe as long as the live coding framework stays 3 levels) — unrelated to this metadata change.

### DIFF
--- a/src/qfa/domain/models.py
+++ b/src/qfa/domain/models.py
@@ -31,6 +31,11 @@ class FeedbackRecordMetadataModel(BaseModel):
     richer metadata (e.g. `fixtures/analyze_corpus.yaml`'s `theme`,
     `language`, `codes`) must project down to these four fields before
     constructing a `FeedbackRecordModel`.
+
+    `feedback_record_id` is a deprecated legacy field carried through from the
+    API model so the boundary mapping doesn't have to strip it; it is ignored
+    (the record-level `id` is the identifier the backend uses). See
+    `ApiFeedbackRecordMetadata` for the full rationale.
     """
 
     model_config = ConfigDict(frozen=True, extra="forbid")
@@ -53,6 +58,16 @@ class FeedbackRecordMetadataModel(BaseModel):
     coding_level_3: str | None = Field(
         default=None,
         description="Code level 3 label assigned to the feedback record.",
+    )
+
+    feedback_record_id: str | None = Field(
+        default=None,
+        deprecated=True,
+        description=(
+            "Deprecated legacy field, accepted for backward compatibility with"
+            " older EspoCRM flowcharts and ignored. The record-level `id` is"
+            " the identifier the backend uses."
+        ),
     )
 
 
--- a/src/qfa/domain/models.py
+++ b/src/qfa/domain/models.py
@@ -31,13 +31,6 @@ class FeedbackRecordMetadataModel(BaseModel):
     richer metadata (e.g. `fixtures/analyze_corpus.yaml`'s `theme`,
     `language`, `codes`) must project down to these four fields before
     constructing a `FeedbackRecordModel`.
-
-    The deprecated `feedback_record_id` key that older EspoCRM flowcharts wrote
-    into metadata is *not* accepted here: it is a wire-compatibility concern
-    stripped at the API boundary (see `ApiFeedbackRecordMetadata` and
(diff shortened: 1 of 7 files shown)


mariushelf:
Keep the field only in the API schema, but not in the domain model. Then in the endpoint, we can simply exclude it from domain model creation. That's exactly what the translation layer in hexagonal architecture is for.

mariushelf:
Done in 727034b. Moved the deprecated `feedback_record_id` off the domain model — it now lives only on `ApiFeedbackRecordMetadata`, and a new `_to_domain_metadata()` helper in `routes.py` (alongside `_to_domain_coding_node`) strips it when translating to the domain `FeedbackRecordMetadataModel`, which is strict (`extra="forbid"`) again.

- `domain/models.py`: field removed; domain shape back to `{created, coding_level_1/2/3}`.
- `api/routes.py`: all 5 mapping sites now go through `_to_domain_metadata` (also DRYs up the repeated `model_validate(model_dump())`).
- `services/prompts.py`: reverted the prompt-exclusion (unnecessary now that the domain model can't hold the key).
- Tests: domain test flipped to assert the key is *rejected* at the domain layer; added a boundary test that the API model accepts it and the mapping drops it; the end-to-end old-EspoCRM-payload → 200 test is unchanged.

`make lint` + `make test` green (526 passed).

src/qfa/domain/models.py
```diff
@@ -31,6 +31,13 @@ class FeedbackRecordMetadataModel(BaseModel):
     richer metadata (e.g. `fixtures/analyze_corpus.yaml`'s `theme`,
     `language`, `codes`) must project down to these four fields before
     constructing a `FeedbackRecordModel`.
+
```
mariushelf: remove this comment. models.py should not change.

mariushelf:
Done in f2f2b15. Removed the explanatory docstring paragraph so `src/qfa/domain/models.py` now has **zero diff vs `main`** — the domain core no longer changes at all. The strip-at-the-boundary behaviour stays entirely in `ApiFeedbackRecordMetadata` + `routes._to_domain_metadata`.

`make lint` green (3 contracts kept); `tests/domain/test_models.py` + `tests/api/test_routes.py` → 83 passed.

PR #235, opened 2026-07-24
fix(espo): fixed small espo prd issues

olafdegraaff:
Updated EspoCRM, resolving following issues:
- empty feedback records in analyse/bulk_summary do not cause errors anymore
- "" are now replaced by - before making the api call
- base_url is now everywhere dynamic and not hard coded
- the summary and detect sensitive do not receive the coding framework anymore in their mother payload, saving tokens


olafdegraaff:
already resolved by #237

PR #243, opened 2026-07-29
fix(analyze): output-quality fixes from first user test

olafdegraaff:
## Summary
Five targeted fixes from feedback during the first round of user testing on `analyze-bulk`/`summarize-bulk` (EspoCRM Insights) and `assign-codes`:

- **Remove the AI disclaimer** ("Generated by AI. Human review required...") from `analyze-bulk` responses — no longer needed. This is already hardcoded in EspoCRM, currently it was double. Daan makes sure the disclaimer in EspoCRM contains all the needed information.
- **`output_language` now always wins**: the directive explicitly overrides both the feedback records' own language and any conflicting language request inside the analyst's free-text prompt, and now also reaches the hierarchical map (per-chunk) step and the judge's `uncertainty_explanation`, not just the final reduce/analysis text.
- **No more trailing questions**: analyze-bulk, summarize-bulk, and single-record summarize no longer close with something like "Would you like me to dig deeper into X?" — added an explicit guardrail against it.
- **Coding-level explanations capped at two sentences**: `assign-codes` builds its combined explanation from one judge explanation per hierarchy level; each is now bounded so the combined result doesn't run long.
- **Dropped feedback record `ids` from the summarize-bulk response** — not needed by the consumer, and was also being requested from (then immediately discarded from) the LLM's structured output.

Each fix updates the relevant architecture docs and adds/updates unit tests alongside the prompt/schema change.

## Test plan
- [x] `make test` — 542 tests pass
- [x] `make lint` — ruff, ty, import-linter all clean
- [ ] Manual smoke test against a real LLM for each of the five behaviors (disclaimer gone, language honored under conflict, no trailing question, short coding explanations, no `ids` in summarize-bulk response)


src/qfa/api/routes.py
```diff
@@ -191,9 +189,9 @@ async def analyze_bulk(
     records = _drop_empty_records(body.feedback_records)
     if not records:
         # All records were empty: nothing to analyze. Return a 200 empty
-        # result (disclaimer preserved) rather than failing the request.
+        # result rather than failing the request.
         return ApiAnalyzeBulkResponse(
-            analysis=ANALYZE_DISCLAIMER,
+            analysis="",
```
olafdegraaff: Now that this is a bare `""` instead of `ANALYZE_DISCLAIMER`, the all-empty-batch response has no explanatory text left in it at all: `_create_pretty_output(quality_score=None, title="Analysis", summary="")` still unconditionally appends the `SUMMARY:` line (`if summary is not None`, and `""` is not `None`), so EspoCRM's rendered `modelResponse` becomes a `TITLE: Analysis` / `SUMMARY:` block with an empty body. `uncertainty_explanation` does carry `_NO_CONTENT_EXPLANATION`, but that field isn't part of `pretty_output`, so the actual text written into EspoCRM never explains why it's blank. Might be worth a fallback string here, or folding the no-content explanation into `pretty_output` for this case.

src/qfa/services/prompts.py
```diff
@@ -176,13 +183,24 @@ def build_analyze_user_message(
...
+    ``uncertainty_explanation`` — the only free text in the judge's response
+    that reaches the analyst — to the same language requested for the
+    analysis itself.
+    """
     return ANALYZE_JUDGE_PROMPT.format(
         source_text=source_text,
         analyst_prompt=analyst_prompt,
         analysis=analysis,
+    ) + build_output_language_instruction(
+        output_language, subject="uncertainty_explanation"
```
olafdegraaff: `subject="uncertainty_explanation"` passes the raw snake_case field name into the shared builder, producing "Write the uncertainty_explanation in Dutch, regardless of..." — every other call site uses human-readable phrasing (`"analysis"`, `"title and summary"`). Consider `subject="uncertainty explanation"` for consistency and slightly better model compliance.

docs/architecture/06-prompt-envelope.md
```diff
@@ -72,11 +98,10 @@ envelope because `<` becomes `&lt;`. The model is instructed explicitly in
 `ANALYZE_GUARDRAILS_PROMPT` to treat envelope content as data.
 
 Note: structural escaping is a depth-in-defence measure, not the primary
-defence. The primary defences are the human-in-the-loop review process and
-the server-prepended `ANALYZE_DISCLAIMER` ("Generated by AI. Human review
-required.") at the top of every analysis response. Regex-based injection
-detection (already present in `LiteLLMClient._check_injection`) and
-future classifier-based detection (tracked in
+defence. The primary defence is the human-in-the-loop review process.
```
olafdegraaff, line 101: This drops the disclaimer from this repo's documented "primary defences" list, leaving human-in-the-loop review as the sole primary defence documented *here*. Per the PR description EspoCRM now owns the disclaimer, so the actual defense-in-depth story likely still holds — it's just enforced by a system outside this repo's tests/visibility now. Worth a quick confirmation with whoever owns that EspoCRM flow (Daan, per the PR description) that it's actually wired before treating this as equivalent coverage.

olafdegraaff (commented):
Self-review — ran through code quality, correctness, docs, and test coverage across all five fixes. Overall the pattern is consistent and well-tested; a few things worth a look before merge, left inline where they anchor to a specific line. Three PR-level notes that don't anchor to a single diff line:

**Missing breaking-change marker on the `ids` removal.** `pyproject.toml` has `[tool.semantic_release]` with `commit_parser = "conventional"`. Removing a field from a public API response (`ApiSummarizeBulkResponse.ids`) is a breaking change and needs `!` in the commit type (e.g. `fix(summarize)!: ...`) to trigger a MAJOR bump — a `BREAKING CHANGE:` footer isn't an option here since AGENTS.md forbids commit trailers. As committed, this releases as a plain PATCH with no changelog signal that `.ids` disappeared, which could silently break any consumer besides EspoCRM reading it directly.

**`AggregateSummaryResultModel.quality_score` (domain/models.py) has the same discarded-LLM-output pattern the PR just fixed for `ids` on this same model** — it's outside this diff's context window so I can't anchor a comment to it directly. The LLM is asked to produce a `quality_score` guess, then `orchestrator.py`'s `summarize_bulk` unconditionally overwrites it with the real judge score (`response.structured.quality_score = quality_score`) before it's ever used. Not a correctness bug — the guessed value never reaches a caller — but it's the identical wasted-generation issue this PR called out for `ids`, left unaddressed on the same model. Possibly worth a follow-up.

**Token-budget creep (informational).** `ANALYZE_GUARDRAILS_PROMPT` and the strengthened `build_output_language_instruction` add fixed overhead to every single-pass/map/reduce/judge system message — confirmed directly, since this PR had to retune `test_recursion_fires_on_corpus_at_least_five_times_token_cap` to compensate. Real corpora already close to `max_total_tokens` could newly tip into `413 payload_too_large` or an extra hierarchical recursion level. Worth a quick post-deploy sanity check.

Verified as non-issues: `assign-codes` has no `output_language` field at all, so it's correctly out of scope for the language-precedence fix; `summarize_bulk`/`summarize`'s judge calls only ever produce a bare float score (never free text), so they correctly needed no language wiring either.

src/qfa/api/routes.py
olafdegraaff: fair point, it is fixed.

src/qfa/services/prompts.py
olafdegraaff: well small thing, but sure I can fix it

docs/architecture/06-prompt-envelope.md
olafdegraaff: It is, trust me Olaf ;)

docs/architecture/06-prompt-envelope.md
```diff
@@ -27,6 +27,32 @@ The orchestrator composes them as:
...
+The directive is explicit about precedence over two failure modes observed in
+practice: the model defaulting to the feedback records' own language, and the
+model instead following a conflicting language request embedded in the
+analyst's free-text `<analyst_instruction>` prompt. Living in the system
+message (never the untrusted user message) is what makes this precedence
+enforceable — see [Selective de-anonymisation](#selective-de-anonymisation-person-retention)
+below for the parallel trusted/untrusted split.
+
+The same directive (via the same builder, with a different `subject`) is
+applied everywhere free text reaches the analyst: the single-pass analyse
```
mariushelf, line 49: This sentence claims the directive is "applied everywhere free text reaches the analyst," then lists four system messages. But `uncertainty_explanation` is also free text reaching the analyst, and on three of its four code paths it's an English string that never sees this builder:

- `mode="hierarchical"` — **always** deterministic English (`"Leaf-judged confidence is a coverage-weighted mean over N of M chunk(s); the lowest single-chunk faithfulness was X."`), assembled in `Orchestrator.analyze_hierarchical`, not by the judge LLM.
- single-pass with a failed judge — `JUDGE_UNAVAILABLE_EXPLANATION`.
- empty batch — `_NO_CONTENT_EXPLANATION`.

So the one mode built for large corpora — the one most likely to hit a mixed-language corpus in the first place — still returns a Dutch analysis with an English explanation.

Keeping server-generated status text in a fixed language is a defensible choice, and I'd rather see it written down than localized. Suggest narrowing the claim to what the code does, e.g.:

> The same directive is applied to every system message that produces analyst-facing **LLM prose**: the single-pass analyse message, the single-pass judge message (pinning `uncertainty_explanation`), and — for `mode=hierarchical` — both the map and reduce messages.
>
> Server-generated explanation text is deliberately **not** localized: the hierarchical `uncertainty_explanation`, `JUDGE_UNAVAILABLE_EXPLANATION`, and `_NO_CONTENT_EXPLANATION` are fixed English strings, so an operator reading logs or a support ticket sees stable, greppable wording regardless of the request's `output_language`.

That turns a doc inaccuracy into a documented design decision, and saves the next reader from re-deriving the asymmetry from the orchestrator.

~Claude

src/qfa/services/prompts.py
```diff
@@ -47,6 +46,9 @@
     "explicitly rather than fabricating support.\n"
     "- Do not produce content that is sensitive, harmful, discriminatory, "
     "or that takes operational action on the analyst's behalf.\n"
+    "- Do not end with a question, an offer of further help, or an "
```
mariushelf, line 49: **Consider:** the same rule is now written three times in two wordings — here, and byte-identically at `orchestrator.py:92` and `orchestrator.py:105`. The tests bind them together only by the shared substring `"Do not end with a question"`, so editing one wording and not the others passes CI while silently splitting the behaviour across endpoints.

Since all three prompts already import from this module, one constant here fixes it:

```python
NO_TRAILING_QUESTION_RULE: str = (
    "Do not end with a question, an offer of further help, or an "
    "invitation for follow-up input."
)
```

…interpolated into the guardrail bullet (keeping its `- ` prefix and the `e.g. "Would you like..."` elaboration) and into both summarization prompts. The tests then assert `NO_TRAILING_QUESTION_RULE in <prompt>` — an identity check that can't drift — instead of a hand-copied substring.

**Separately, and lower conviction:** `docs/architecture/06-prompt-envelope.md` describes `ANALYZE_GUARDRAILS_PROMPT` as *"Hard rules that must be obeyed regardless of any other content"* and justifies the three-constant split as keeping guardrail text *"auditable at a glance."* Every other bullet is safety-critical (data-not-instructions, no identification, no fabrication); this one is a tone preference. Your reasoning for putting it here is sound and I'd have done the same — it's the only constant shared by single-pass, map, and reduce. If you want the audit property back without losing that reach, a sibling `ANALYZE_STYLE_PROMPT` composed alongside the guardrails in those same three builders gets both. Fine to leave as-is; if you do, the doc table row for `ANALYZE_GUARDRAILS_PROMPT` should mention it now carries output-style rules too.

~Claude

src/qfa/services/orchestrator.py
```diff
@@ -102,6 +102,7 @@
     "Scale the number of bullet points to the size and diversity of the input — use judgement.\n"
     "Also create a short, 3-5 word descriptive title reflecting the dominant theme.\n"
     "Do not include markdown code fences.\n"
+    "Do not end with a question, an offer of further help, or an invitation for follow-up input.\n"
     "Use the same language as the input feedback records unless a target language is specified."
```
mariushelf, line 106: **Nit:** this system message can now carry two language rules that argue with each other. This line says *"Use the same language as the input feedback records unless a target language is specified"*; when `output_language` is set, `build_output_language_instruction` appends *"…regardless of the language the feedback records are written in. This directive takes precedence over any other language request."*

The `unless a target language is specified` clause does resolve it logically, so this isn't a bug. But the whole point of the strengthened wording is that models were latching onto the wrong language cue, and this line is a competing cue sitting a few tokens above the directive. Since `build_output_language_instruction` already handles the set case and returns `""` for the unset case, the fallback intent could live entirely in the builder — leaving exactly one language rule in the prompt at any time.

Same applies to `_DEFAULT_SUMMARIZATION_PROMPT` at line 93, though `/v1/summarize` has no `output_language` input at all, so *"unless a target language is specified"* is unreachable there today.

~Claude

tests/services/test_orchestrator.py
```diff
@@ -733,6 +765,85 @@ async def test_no_language_directive_when_output_language_unset(self, settings):
         )
 
 
+class TestNoTrailingQuestion:
```
mariushelf, line 768: None of the three test methods in this class has a docstring — every other test added or touched in this PR does, including the ones right above in `TestOutputLanguage` and the new `test_coding_classifier.py`. The class docstring carries the *why*, which is the valuable half, but the per-test *what* is missing, so a failure reports only a method name.

One line each is enough, e.g.:

```python
async def test_summarize_bulk_system_message_forbids_trailing_questions(self, settings):
    """The aggregate-summarization prompt carries the no-trailing-question rule."""
```

**Consider**, while you're here: all three tests construct a full `Orchestrator` + `FakeLLMPort` and drive a complete two-call round trip in order to assert a substring is present in `calls[0]["system_message"]`. That's a lot of setup to verify a constant's contents, and it re-tests the composition logic already covered elsewhere. `test_prompts.py::test_guardrails_forbid_trailing_questions` asserts the same property against the constant directly, in three lines and no fixtures.

The orchestrator-level tests do earn their keep for the two `_DEFAULT_*_SUMMARIZATION_PROMPT` constants, since those aren't otherwise proven to reach the LLM. But `test_analyze_bulk_system_message_forbids_trailing_questions` is the composition of two things already independently tested. Either drop it, or keep it and note in the docstring that it's guarding *composition* (the guardrails actually reach the analyse call) rather than *content* — otherwise the next reader sees three near-identical tests and can't tell which invariant each one owns.

~Claude

tests/services/test_orchestrator_hierarchical.py
```diff
@@ -239,10 +243,13 @@ async def test_anonymization_happens_before_any_llm_or_embed_call():
 class LargeOutputLLM:
     """LLM that returns a moderate map output to force multi-level tree-reduce.
 
-    Map calls return a ~480-char partial. With a 700-token budget and a
-    393-token system message, exactly 2 partials fit per reduce group
-    (393+120+120=633 ≤ 700; 393+120+120+120=753 > 700). Eight map chunks
-    therefore produce a tree-reduce of depth ≥ 2 (4 groups → 2 groups → 1).
+    Map calls return a ~250-char partial. With a 700-token budget and a
+    442-token system message, 3 partials (plus envelope overhead) fit in one
```
mariushelf, line 247: This is the finding I'd most want addressed in the tests, because it will keep costing you.

Adding one line to `ANALYZE_GUARDRAILS_PROMPT` forced retuning four coupled magic numbers here: partial size 480→250 chars, chunk count 8→10, the system-message token count 393→442, and the resulting fan-out 2-per-group→3-per-group — plus the matching prose in two docstrings. The test's actual subject ("recursion fires on a corpus ≥5× the token cap") is unrelated to any of that; it broke purely because an unrelated prompt got one line longer, and the docstring now has to say *"measured empirically … since the per-partial envelope tags make the token count not a clean multiple."*

That's a test pinned to an implementation detail that changes for reasons the test doesn't care about — and prompt text will keep changing. The recursion property is worth testing; the arithmetic isn't.

Cheapest fix: derive the threshold instead of hardcoding it, so the numbers follow the prompt automatically:

```python
_SYSTEM_TOKENS = _count_tokens(build_reduce_system_message())
_BUDGET = _SYSTEM_TOKENS + 3 * _PARTIAL_TOKENS + _ENVELOPE_SLACK  # 3 partials fit, 4 don't
```

…passing `_BUDGET` as `max_total_tokens` rather than the literal 700. The test then asserts what it means — *a corpus this much larger than the budget forces a multi-level tree-reduce* — and survives the next prompt edit untouched. If deriving it is more machinery than it's worth, even hoisting the four numbers into named module constants next to a one-line comment ("retune if the reduce system prompt changes size") would beat having them spread across two docstrings and two inline comments.

~Claude

mariushelf (commented):
## Review: output-quality fixes from first user test

Five-axis review (correctness, readability, architecture, security, performance). I read the full diff, verified the claims against the branch, and re-ran the verification story.

**Verification I ran myself** on `625ae74`:
- `uv run pytest -q` → **542 passed**, 42 deselected. Matches the PR description.
- `make lint` → clean; import-linter **3/3 contracts kept** (hexagonal layers, domain-free-of-infrastructure, services-depend-only-on-ports).
- Confirmed `ids` is fully gone — no dangling references in `src/`, `docs/`, or the EspoCRM integration notes. The removal is clean end to end (domain model → API schema → `pretty_output` → orchestrator write-back → structured-output schema). Dropping a field the LLM was asked to produce and then immediately overwrote server-side is a genuine simplification, not just a relocation.
- Confirmed the `_judge_chunk` comment is **accurate**: in hierarchical mode `uncertainty_explanation` really is a deterministic f-string built from aggregated scores, so skipping `output_language` on the leaf judge costs nothing. Good comment — it documents a non-obvious asymmetry that would otherwise read as an oversight.

The three items from your self-review (empty-batch fallback text, `subject="uncertainty explanation"` wording, EspoCRM disclaimer ownership) are all addressed on the current head, so I've skipped them.

Overall this is well-scoped and consistently tested, and each of the five fixes lands where it should. Two things I'd want resolved before merge — both are about the change being *described* accurately rather than the code being wrong.

---

### 1. This is a breaking API change shipping as a patch release

Removing `ids` from the `POST /v1/summarize-bulk` 200 response is a breaking change for any client reading that field. Three places don't reflect that:

- **Release versioning.** `pyproject.toml` configures `python-semantic-release` with `commit_parser = "conventional"`. A `fix(analyze):` subject cuts **2.1.1** — a patch release that silently removes a response field.
- **`docs/rest-api/index.md` § Breaking changes.** That section exists and records the 0.14.0 field rename; this change gets no entry.
- **`docs/integrations/espo-crm.md`** documents the deploy-independence contract: *"a new backend can be deployed without importing updated flowcharts first, and vice versa."* An older flowchart that reads `ids` breaks against a new backend, which is exactly the direction that doc promises is safe.

I take the PR description at its word that the consumer doesn't use `ids`. The ask isn't to keep the field — it's to signal the removal. Note `AGENTS.md` bans commit trailers of any kind, so a `BREAKING CHANGE:` footer is out; the convention-compliant signal is the bang in the type, e.g. `fix(analyze)!: …`. Plus a line under § Breaking changes.

### 2. `output_language` reaches the analysis, but not most of the explanation paths

The fix genuinely works for the happy path of single-pass mode. But `uncertainty_explanation` is analyst-facing free text on **four** paths, and only one of them now honours the directive:

| Path | `uncertainty_explanation` source | Honours `output_language`? |
|---|---|---|
| single-pass, judge OK | LLM (judge) | ✅ this PR |
| single-pass, judge failed | `JUDGE_UNAVAILABLE_EXPLANATION` | ❌ English constant |
| **hierarchical (always)** | deterministic f-string in `analyze_hierarchical` | ❌ English |
| empty batch | `_NO_CONTENT_EXPLANATION` | ❌ English constant |

So `output_language="Dutch"` + `mode="hierarchical"` returns a Dutch analysis alongside *"Leaf-judged confidence is a coverage-weighted mean over 3 of 4 chunk(s); the lowest single-chunk faithfulness was 0.42."* — the mixed-language output this PR set out to eliminate.

I don't think the code must change: server-generated status text arguably belongs in a stable operator language, and localizing it means either an LLM round-trip or a translation table. That's a reasonable product call. But it should be a *stated* call — see the inline comment on the new docs section, which currently claims the directive is applied "everywhere free text reaches the analyst."

### 3. On what the new tests can and can't prove

Four of the five fixes are prompt-only. The new tests assert that a constant contains a substring — which proves the constant was edited, not that the model complies. That makes them useful change-detectors and honest regression guards, and I'd keep them; it just means the unchecked *"Manual smoke test against a real LLM"* box in the test plan is carrying the actual verification weight for the disclaimer removal, the language-precedence wording, the no-trailing-question guardrail, and the two-sentence cap. Worth ticking before merge rather than after.

---

**Verdict: approve once (1) and (2) are addressed.** Nothing here is functionally broken, the architecture is respected, and no security or performance concern surfaced — removing the disclaimer moves an AI-provenance signal from the API to EspoCRM, which you've confirmed with the owner. Remaining inline comments are `Consider`/`Nit` and the author's call.

~Claude

mariushelf (approved):
LGTM. The uncertainty explanation is still hardcoded to English at least in some places, but that can either be acceptable or solved in a follow-up.

PR #244, opened 2026-07-30
fix(coding): surface near-miss explanation when threshold rejects eve…

### LINKED ISSUE #242: [BUG] empty array assign codes
Sometimes the assign codes gives back an empty array []. There is an example in prd were this happened (ask Daan). It is still unclear why this sometimes happen.

- Figure out what causes is
- Fix it

olafdegraaff:
Closes #242 
Previously, when confidence_threshold filtered out every candidate at every hierarchy level, assign-codes returned an unexplained empty assigned_codes list, indistinguishable from a malformed LLM response or a genuine "no code applies" pick. Now it returns a single entry with null coding_level_*/confidence_* fields and an explanation describing the highest-scoring rejected candidate.

### DIFF
--- a/docs/rest-api/index.md
+++ b/docs/rest-api/index.md
@@ -60,6 +60,8 @@ Per-record inference endpoints (`/v1/summarize`, `/v1/assign-codes`, `/v1/detect
 
 Empty `content` is accepted on every endpoint and never causes a 422. A record with empty content carries no information, so it is dropped (bulk) or short-circuited to a 200 empty result with no LLM call (per-record) — see each endpoint's reference for the exact empty-result shape. This keeps a single blank EspoCRM description from silently failing a whole request (issue #138).
 
+`POST /v1/assign-codes` accepts an optional `confidence_threshold`; codes whose judge score falls below it are filtered out. If every candidate at every hierarchy level is filtered out this way, the response is a 200 with a single `assigned_codes` entry whose `coding_level_*`/`confidence_*` fields are `null` and whose `explanation` describes the highest-scoring rejected candidate — instead of an unexplained empty list.
+
 ## Usage endpoint response shape
 
 All usage endpoints return aggregated stats in two parallel views:
--- a/docs/rest-api/index.md
+++ b/docs/rest-api/index.md
@@ -60,7 +60,7 @@ Per-record inference endpoints (`/v1/summarize`, `/v1/assign-codes`, `/v1/detect
 
 Empty `content` is accepted on every endpoint and never causes a 422. A record with empty content carries no information, so it is dropped (bulk) or short-circuited to a 200 empty result with no LLM call (per-record) — see each endpoint's reference for the exact empty-result shape. This keeps a single blank EspoCRM description from silently failing a whole request (issue #138).
 
-`POST /v1/assign-codes` accepts an optional `confidence_threshold`; codes whose judge score falls below it are filtered out. If every candidate at every hierarchy level is filtered out this way, the response is a 200 with a single `assigned_codes` entry whose `coding_level_*`/`confidence_*` fields are `null` and whose `explanation` describes the highest-scoring rejected candidate — instead of an unexplained empty list.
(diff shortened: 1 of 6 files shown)


olafdegraaff (commented):
## Overview

This PR changes `POST /v1/assign-codes` behavior for the case where `confidence_threshold` filters out every candidate at every hierarchy level. Previously that case returned an unexplained empty `assigned_codes` list — indistinguishable from a malformed LLM pick response or a genuine "no code applies" decision. Now it returns a single `assigned_codes` entry with `coding_level_*`/`confidence_*` fields set to `null` and `explanation` describing the highest-scoring rejected candidate ("near-miss").

To support this, `coding_level_1_id`, `coding_level_1_name`, `confidence_level_1`, and `confidence_aggregate` move from required to optional (nullable) on both `AssignedCodeModel` (domain) and `ApiAssignedCode` (API schema). The orchestrator tracks the best-scoring rejected candidate across the whole recursive traversal via a `best_rejected` out-parameter threaded alongside the existing `candidates` accumulator.

## Code Quality & Style

- The change is well-scoped and follows the existing `_ScoredCode`/`candidates` accumulator pattern for the new `best_rejected` tracking — consistent with how the recursive traversal already works, rather than introducing a new mechanism.
- Docstrings are updated in three places (`routes.py`, `orchestrator.py`, and the new field descriptions in both schema files) — good for keeping the contract self-documenting, and `docs/rest-api/index.md` gets a matching note.
- Moving `new_path`/`new_scores`/`new_explanations` computation above the threshold check (`orchestrator.py`) so both the accepted and rejected branches can reuse them is a clean way to avoid duplicating the accumulation logic.

## Specific Suggestions

1. **Redundant guard** — `src/qfa/services/orchestrator.py`, the `elif not candidates and best_rejected[0] is not None:` branch. `CodingAssignmentRequestModel.max_codes` is constrained to `ge=1`, so `top = candidates[:max_codes]` is falsy if and only if `candidates` is already empty. The `not candidates` check is therefore always true whenever this `elif` is reached and can be simplified to `elif best_rejected[0] is not None:`. Not a bug, just a needless condition — a future reader may wonder what scenario the guard is protecting against.

2. **API design note** — the near-miss entry and a genuine leaf-code assignment now share the exact same `ApiAssignedCode` shape inside the same `assigned_codes` array, distinguished only by `coding_level_1_id is None`. Any existing consumer that treats `len(assigned_codes) == 0` as "nothing was coded" will now see a length-1 array in the threshold-rejection case and needs to additionally check for the null sentinel. This is presumably intentional (the requested design was "nullable entry in the array" over "new top-level field"), but worth being explicit about in the docs paragraph since it's a subtle behavior change for integrations built against the old contract — e.g. EspoCRM-style consumers polling this endpoint.

## Potential Issues & Risks

- **Test coverage gap (HTTP layer)**: `tests/api/conftest.py`'s `FakeOrchestrator.assign_codes` always returns a fully-populated `AssignedCodeModel` regardless of `confidence_threshold`, so `tests/api/test_routes.py` never exercises the new null-field shape through actual FastAPI response validation/serialization (`ApiAssignedCode` with `coding_level_1_id=None` etc.). The three new tests in `tests/services/test_orchestrator.py` cover the orchestrator/domain logic well, but there's no assertion that a `None` domain value actually round-trips to JSON `null` via the route handler's response model. Given this PR changes required→optional on a public API schema, a route-level test (or an update to `FakeOrchestrator` to support this scenario) would close the loop.
- **Test coverage gap (nested rejection)**: all three new tests reject at the root level (single-level or two-sibling root trees). None exercises a rejection that happens after descending into a child level, which goes through the same `_traverse_coding_level` recursive call but with a non-empty `path_ids_names`/`accumulated_scores` — the main other way to reach the `best_rejected` assignment. Given the recursive nature of this function, a nested case is worth one test to guard against regressions in path/score accumulation for the rejected branch specifically.
- **Breaking change is schema-only (no explicit call-out beyond docs)**: since `coding_level_1_id`/`name`/`confidence_level_1`/`confidence_aggregate` go from required to optional, strongly-typed clients (generated from the OpenAPI schema) could break on receiving `null` where a string/float was previously guaranteed. Version bumping appears automated via `python-semantic-release` off the conventional-commit type, so no manual action needed there — just flagging that this is a real (if narrow) breaking change for typed consumers, beyond the additive "empty → populated" changes this repo's `docs/migration/` has documented before.

Nothing here blocks merging — the core logic is sound, well-tested at the orchestrator level, and the docs are updated. The main asks are: simplify the redundant guard, and consider adding one HTTP-level test (or a nested-rejection unit test) to cover the layer that isn't currently exercised.

mariushelf (approved):
Interesting approach. List of "candidates", but without coding levels, just explanation. LGTM.

PR #249, opened 2026-08-05
Feat/monitoring2.0

olafdegraaff:
This solve part of the monitoring issues, but it is not quite fixed yet. It is a start.

## Summary

Makes container logs actually reach Azure, and adds a saved query for browsing API calls.

- **Enable Application Logging (Filesystem)** on the App Service (`infra/app_service.tf`) — without it, the container's stdout/stderr was never captured at all, so nothing reached the Log stream or the diagnostic setting regardless of what the app logged.
- **`PYTHONUNBUFFERED=1`** in the `Dockerfile` — Python block-buffers stdout/stderr when it isn't attached to a TTY (always true in a container), so log lines could sit in a buffer instead of reaching the log pipeline promptly.
- **Force our logging handler in `setup_logging()`** (`src/qfa/utils.py`) — `configure_azure_monitor()` attaches its own handler to the root logger before `setup_logging()` runs in every deployed environment, which made the plain `basicConfig()` call a silent no-op.
- **Saved KQL query** `qfa-<env>-api-calls-overview` (`infra/observability.tf`) against `AppServiceHTTPLogs`, so "what API calls have been made and what did they return" doesn't require hand-writing KQL each time.
- **Docs**: document the Application Logging dependency and the new saved query in `docs/operations/observability.md`.

## Test plan
- [ ] `make test` / `make lint`
- [ ] `terraform plan` against a real environment to confirm the `application_logs` and saved-search resources apply cleanly
- [ ] Deploy and confirm log lines show up in the App Service Log stream / `AppServiceConsoleLogs`
- [ ] See review comment re: verifying `AppTraces` (Application Insights) still receives log lines after the `force=True` change

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -36,6 +36,12 @@ ARG EMBEDDING_E5_REVISION
 COPY --from=ghcr.io/astral-sh/uv:latest /uv /usr/local/bin/uv
 WORKDIR /app
 
+# Force stdout/stderr unbuffered. Without this, Python block-buffers output
+# that isn't attached to a TTY (always true in a container), so log lines can
+# sit in a buffer indefinitely instead of reaching the App Service log
+# pipeline in anything like real time.
+ENV PYTHONUNBUFFERED=1
+
 # Model layer first: it's large and rarely changes, so keeping it above the
 # frequently-churning app code means it stays cached (and isn't re-pushed)
 # until the pinned revision changes. Only the default model (e5-base) is
--- a/docs/operations/observability.md
+++ b/docs/operations/observability.md
@@ -4,9 +4,11 @@ Logging, request tracing, and usage reporting.
 
 ## Where logs go
 
-The app writes structured logs to stdout. On Azure App Service these are picked up by the App Service log stream and any attached log sink. Locally, they show up in your terminal.
+The app writes structured logs to stdout/stderr. On Azure App Service these are picked up by the App Service log stream and any attached log sink. Locally, they show up in your terminal.
 
-The log level is controlled by `LOG_LOGLEVEL` for application packages (default `DEBUG`) and `LOG_LOGLEVEL_3RDPARTY` for third-party libraries (default `WARNING`). Both accept either a numeric level or a string (`"debug"`, `"info"`, `"warning"`, `"error"`, `"critical"`).
+Capture depends on **Application Logging (Filesystem)** being on for the App Service — this is the `application_logs { file_system_level = "Information" }` block on `azurerm_linux_web_app.backend.logs` in `infra/app_service.tf`. This is a separate setting from `http_logs` (raw per-request access log) and from the diagnostic setting in `infra/observability.tf` (which ships both into Log Analytics) — without it, the container's stdout/stderr is never captured at all, so there's nothing in the Log stream and nothing for the diagnostic setting to route, regardless of what the app logs or how the diagnostic setting is configured.
+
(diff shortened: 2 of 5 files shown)


olafdegraaff:
## Review

### Overview
This PR closes the loop on container logging reaching Azure: it turns on **Application Logging (Filesystem)** on the App Service (`infra/app_service.tf`), unbuffers Python's stdout/stderr in the container (`Dockerfile`), forces our own logging handler to survive `configure_azure_monitor()` (`src/qfa/utils.py`), adds a saved KQL query for browsing API calls (`infra/observability.tf`), and documents all of it (`docs/operations/observability.md`).

Each individual diagnosis in the comments is correct and well explained. One change has a side effect that undoes part of the existing telemetry setup — see below.

### 🔴 Likely regression: `force=True` silently kills the Application Insights log pipeline
`src/qfa/utils.py`:
```python
logging.basicConfig(
    level=log_config.loglevel_3rdparty, force=True, **log_config.basicConfig
)
```
Per the Python docs, `force=True` **removes and closes every existing handler on the root logger** before adding the new one. `qfa.main` calls `configure_azure_monitor()` *before* `setup_logging()` runs whenever `APPLICATIONINSIGHTS_CONNECTION_STRING` is set (i.e. every deployed environment) — and that call attaches an OpenTelemetry `LoggingHandler` to the **root logger** (verified in `azure/monitor/opentelemetry/_configure.py::_setup_logging`, `logger_name` defaults to `""`).

So in every deployed environment, `setup_logging()` now removes that OTel handler and replaces it with only a `StreamHandler`. I confirmed this locally:
```python
root.addHandler(FakeOtelHandler())
logging.basicConfig(level=logging.WARNING, force=True, format='%(message)s')
# root.handlers is now [StreamHandler], the OTel handler is gone
```
This fixes the immediate bug (logs reaching stdout) but as a side effect permanently severs `AppTraces` delivery for the process lifetime — the app's log lines will no longer reach Application Insights via OTel at all. That directly contradicts the doc line this same PR leaves in place at `docs/operations/observability.md:328-330`:

> Once telemetry is confirmed, your application log lines appear in **both** `AppServiceConsoleLogs` (via the diagnostic setting) and `AppTraces` (via App Insights) — the same lines in two tables, reaching them by two independent paths.

That statement will be false after this change ships.

**Suggested fix:** don't wipe existing handlers — add a `StreamHandler` alongside whatever `configure_azure_monitor()` already attached, e.g.:
```python
handler = logging.StreamHandler()
handler.setFormatter(logging.Formatter(log_config.basicConfig["format"]))
logging.getLogger().addHandler(handler)
logging.getLogger().setLevel(log_config.loglevel_3rdparty)
```
or only pass `force=True` when no handler is present yet, so local/test runs (which don't call `configure_azure_monitor()`) keep today's `basicConfig` behavior unchanged.

### Test coverage
`tests/test_utils.py` has no test for `setup_logging()` at all, so this regression (and the original no-op bug it fixes) wouldn't be caught either way. Worth a test that simulates a handler already present on the root logger (standing in for the OTel handler) and asserts it's still attached — and still receiving records — after `setup_logging()` runs.

### Everything else — looks good
- `Dockerfile`: `PYTHONUNBUFFERED=1` is the standard, correct fix for block-buffered stdout in a container; comment is accurate.
- `infra/app_service.tf`: `application_logs { file_system_level = "Information" }` is the right knob and the comment correctly explains why nothing was captured before. Minor: unlike the `http_logs.file_system` block just below it, there's no explicit retention here — worth a one-line note (or just accept Azure's platform-enforced ~12h rolling window for filesystem app logs) so a reader doesn't wonder why it's asymmetric with `http_logs`.
- `infra/observability.tf`: the saved search is a nice quality-of-life addition; no concerns.
- `docs/operations/observability.md`: the new Application Logging paragraph and the saved-query section are clear and match the infra changes — just needs a pass once the `force=True` issue above is resolved, since it currently documents delivery-to-both-sinks behavior that this PR's logging change removes.

### Risk / rollout
Given the App Insights regression, I'd hold off merging until the `force=True` behavior is fixed (or confirm intentionally accepting the AppTraces log loss and update the docs accordingly) — that's a meaningful step back in observability for a PR whose entire purpose is improving observability.

PR #260, opened 2026-08-07
assign codes: always explain why no code

### LINKED ISSUE #256: assign-codes: always explain why no code was applied
## Spec

**What:** Make `POST /v1/assign-codes` always return at least one `assigned_codes` entry
when no code is applied, carrying a human-readable explanation that leads with
`NO CODING APPLIED.`

Three code paths currently end with no code assigned, and only one of them explains itself:

| # | Trigger | Today returns | Explanation? |
|---|---|---|---|
| 1 | Empty `content` — `src/qfa/api/routes.py:424` (issue #138) | `[]` | none |
| 2 | Every candidate below `confidence_threshold` — `src/qfa/services/orchestrator.py:1367` | one null entry | combined rejections, unlabelled |
| 3 | Nothing selected at any level — `src/qfa/services/orchestrator.py:1374` | `[]` | none |

Cases 1 and 3 leave EspoCRM with nothing at all to display: `4_1_assign_codes.php` reads
`assigned_codes.0.explanation`, gets null, and still sets `autoCodingStatus = 'completed'`.
Case 2 does deliver text, but it is unlabelled and hard to read.

This ticket unifies all three behind one contract and reformats the message.

**Why:** Users see codes silently missing in EspoCRM with no indication why (parent #255).
Case 2's explanation already reaches Espo via `autoCodingExplanation`, so improving that text
is immediately visible to users without any EspoCRM change. Cases 1 and 3 send nothing, so no
amount of formatting helps until they also return an entry.

### Target format (case 2)

```
NO CODING APPLIED.
No code reached the 10% confidence threshold, so this record needs human review.

Shelter > Repairs > Roofing — 4%
  No mention of roof damage; the feedback concerns rent costs.

Water — 3%
  No reference to water access.

Food Security — 2%
  Concerns housing, not food.

5 further codes scored below 2%.
```

Cases 1 and 3 use the same `NO CODING APPLIED.` lead with a case-specific second line and no
candidate list.

### Notes for the implementer

- `_combine_rejected_explanations` (`orchestrator.py:265`) currently takes only `rejected`.
  The threshold is available at the call site (`orchestrator.py:1370`) and must be threaded in.
- A rejected candidate has exactly one sub-threshold level — the traversal `continue`s on the
  first failure (`orchestrator.py:1503-1508`). So the last accumulated score is always the
  minimum, and the decisive level is unambiguous.
- When `confidence_threshold` is `None` nothing is ever rejected, so case 2 cannot occur
  without a real threshold value. The percentage in the lead line always has a value.
- English only. `ApiAssignCodesRequest` has no `output_language` field and this ticket does not
  add one; localization is explicitly out of scope.
- Keep the message in the existing `explanation` string. Do **not** add a `pretty_output`
  property to `ApiAssignCodesResponse` — that would require a matching EspoCRM script change
  and is out of scope here.

## Acceptance criteria

- [ ] `assigned_codes` is never an empty list; when no code is applied it contains exactly one
      entry with null `coding_level_*` / `confidence_*` fields and a non-empty `explanation`
- [ ] Every no-code `explanation` begins with the literal line `NO CODING APPLIED.`
- [ ] Case 1 (empty `content`) yields an explanation stating the feedback text was empty;
      `tests/api/test_routes.py::test_assign_codes_empty_content_returns_no_codes` is updated
      from asserting `== []` to the new shape
- [ ] Case 2 (below threshold) yields a lead line naming the threshold as a whole percentage
- [ ] Case 3 (no candidate selected) yields an explanation stating that no code in the
      framework was judged relevant
- [ ] Confidence values render as whole percentages, not `0.00`–`1.00` decimals
- [ ] Each listed candidate shows only its decisive (final, lowest-scoring) level's judge
      explanation, not one line per hierarchy level
- [ ] At most 3 rejected candidates are listed, highest-scoring first, followed by a single
      line counting the remainder; that count line is absent when 3 or fewer were rejected
- [ ] `_combine_rejected_explanations` receives the confidence threshold from its call site
- [ ] Docstrings in `routes.py` and `orchestrator.py` describing the #138 empty-list contract
      are reworded to match the new behaviour

mariushelf:
## What changed

Three code paths in `assign-codes` could end with no code assigned, and only one of them explained itself. The other two returned a bare `assigned_codes: []`, which EspoCRM surfaced as an empty `autoCodingExplanation` on a record still marked `completed` — a successful call that told the user nothing.

`assigned_codes` is now **never empty**. Whenever no code is applied the response is a 200 carrying exactly one entry with null `coding_level_*` / `confidence_*` fields and an `explanation` whose first line is the literal `NO CODING APPLIED.`

| Trigger | Before | After |
|---|---|---|
| Empty `content` (`qfa.api.routes.assign_codes`) | `[]` | entry: *"The feedback text was empty, so there was nothing to code."* |
| All candidates below `confidence_threshold` | one entry, unlabelled decimal dump | reformatted near-miss list (below) |
| Nothing selected at any hierarchy level | `[]` | entry: *"No code in the framework was judged relevant to this feedback."* |

The below-threshold message was also reformatted for a non-technical reader:

- confidences render as whole percentages (`4%`) instead of raw decimals (`0.04`);
- only the **decisive** level's judge reasoning is shown — the traversal stops at the first sub-threshold level, so the last accumulated level is both the lowest-scoring one and the reason the candidate was dropped. Earlier levels passed and were only noise;
- at most **3** near misses are spelled out, highest-scoring first, with any remainder collapsed into a single count line.

The discriminator for clients is `coding_level_1_id is None`, not list length.

## Examples of the explanations produced

**1. Empty feedback text** — short-circuited with no LLM call (issue #138):

```text
NO CODING APPLIED.
The feedback text was empty, so there was nothing to code.
```

**2. Everything filtered out by `confidence_threshold`** — this exact rendering is pinned by `test_message_matches_the_documented_layout` (threshold `0.1`, eight rejected candidates):

```text
NO CODING APPLIED.
No code reached the 10% confidence threshold, so this record needs human review.

Shelter > Repairs > Roofing — 4%
  No mention of roof damage; the feedback concerns rent costs.

Water — 3%
  No reference to water access.

Food Security — 2%
  Concerns housing, not food.

5 further codes scored below 2%.
```

The trailing count line is cut off at the **lowest listed** candidate's score, so "below 2%" refers to `Food Security — 2%`, the third and last one shown.

**3. Nothing judged relevant at any level** — the LLM selected nothing, so no threshold rejection ever happened:

```text
NO CODING APPLIED.
No code in the framework was judged relevant to this feedback.
```

**Formatting variants of case 2**, each with its own test:

- **Three or fewer near misses** → the remainder line is omitted entirely. Its absence is what tells the reader nothing was truncated; a `0 further codes` line would be noise.
- **Exactly one extra** → singular: `1 further code scored below 7%.`
- **Ordering** → candidates are sorted by score descending, because the traversal appends rejections in framework order, not score order. The closest near miss is always read first.

## Files

- `src/qfa/services/orchestrator.py` — public `NO_CODING_LEAD`, `NO_CODING_EMPTY_CONTENT_EXPLANATION`, `NO_CODING_NOTHING_RELEVANT_EXPLANATION` constants; `_as_whole_percentage`; rewritten `_combine_rejected_explanations(rejected, threshold)`; `_ScoredCode.decisive_explanation`; non-empty `else` branch
- `src/qfa/api/routes.py` — empty-content short-circuit returns an explained entry; route docstring rewritten (drives OpenAPI)
- `src/qfa/api/schemas.py` — `ApiAssignedCode.explanation` description updated
- `tests/services/test_orchestrator.py`, `tests/api/test_routes.py` — 6 new formatter tests + 2 rewritten behaviour tests
- `docs/rest-api/index.md`, `docs/integrations/espo-crm.md` — contract + rendered example
- `.gitignore` — ignore `graphify-out/` build artifacts

## Design notes

- **No `assert` for the threshold narrowing.** A rejection cannot happen without a threshold, which invites `assert threshold is not None`. There is no precedent for `assert` in `src/qfa/` (asserts vanish under `python -O`), so the check is folded into the branch condition — `elif rejected and request.confidence_threshold is not None:` — which satisfies `ty` and degrades to the "nothing relevant" message in the impossible case. No branch is unreachable.
- **The exact-string test is the contract.** Asserting the whole rendered block rather than a few `in` substring checks is what actually pins the blank-line separators, the two-space indentation and the em-dash headers. Substring assertions would pass on a subtly reordered message.
- **Constant placement is driven by `import-linter`.** Cases 1–3 straddle `qfa.api` and `qfa.services`. Publishing the strings from `qfa.services.orchestrator` and importing them into the route keeps one source of truth while flowing with the allowed dependency direction; defining them in the route would force `qfa.services` → `qfa.api`, which the "Enforce hexagonal layers" contract rejects.

## Verification

`make test` 559 passed · `make lint` clean, all 3 `import-linter` contracts kept · `make docs` build succeeded.

## Impact & risk

- **Behavioural change for any client that treats `assigned_codes == []` as "no code".** The list is now never empty; the discriminator is `coding_level_1_id is None`. The known consumer, `scripts/espo_crm/feedback_trigger/4_1_assign_codes.php`, already reads `assigned_codes.0.*` and assigns nulls to the coding-level fields, so it needs **no change** — it simply starts populating `autoCodingExplanation` in the two cases that previously sent nothing. Espo already sends `confidence_threshold: 0.1`, matching the example above.
- Case 1 still short-circuits with **no LLM call**, so there is no cost change. Cases 2 and 3 are pure formatting of data already computed.
- Explanations are **English only** regardless of the feedback language — `ApiAssignCodesRequest` has no `output_language` field, and localization was out of scope for the spec.
- Fully revertible with `git revert`; no schema, auth, or migration changes.

Fixes #256

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -200,3 +200,5 @@ config
 # See https://github.com/mariushelf/copier-devcontainer
 /.devcontainer/
 /.envrc
+.graphify-out
+
--- a/.gitignore
+++ b/.gitignore
@@ -200,5 +200,5 @@ config
 # See https://github.com/mariushelf/copier-devcontainer
 /.devcontainer/
 /.envrc
-.graphify-out
+graphify-out/
 
--- a/docs/rest-api/index.md
+++ b/docs/rest-api/index.md
@@ -59,9 +59,26 @@ the single-pass response too.
 
 Per-record inference endpoints (`/v1/summarize`, `/v1/assign-codes`, `/v1/detect-sensitive`) accept a single `feedback_record` and return one result object, unlike bulk endpoints that accept multiple records and return aggregated output.
 
-Empty `content` is accepted on every endpoint and never causes a 422. A record with empty content carries no information, so it is dropped (bulk) or short-circuited to a 200 empty result with no LLM call (per-record) — see each endpoint's reference for the exact empty-result shape. This keeps a single blank EspoCRM description from silently failing a whole request (issue #138).
+Empty `content` is accepted on every endpoint and never causes a 422. A record with empty content carries no information, so it is dropped (bulk) or short-circuited to a 200 with no LLM call (per-record) — see each endpoint's reference for the exact result shape. This keeps a single blank EspoCRM description from silently failing a whole request (issue #138).
 
-`POST /v1/assign-codes` accepts an optional `confidence_threshold`; codes whose judge score falls below it are filtered out. If every candidate at every hierarchy level is filtered out this way, the response is a 200 with a single `assigned_codes` entry whose `coding_level_*`/`confidence_*` fields are `null` and whose `explanation` combines every rejected candidate's explanation (highest-scoring first) — instead of an unexplained empty list.
+`POST /v1/assign-codes` accepts an optional `confidence_threshold`; codes whose judge score falls below it are filtered out. Its `assigned_codes` list is never empty: whenever no code is applied — because the content was empty, because no candidate reached the threshold, or because nothing in the framework was judged relevant — the response is a 200 with exactly one entry whose `coding_level_*`/`confidence_*` fields are `null` and whose `explanation` begins with the line `NO CODING APPLIED.` followed by the reason. A client therefore always has something to show the user instead of an unexplained empty list.
+
+When the threshold is what blocked coding, the explanation names the threshold as a percentage and lists the closest near misses (at most three, highest-scoring first, each with its decisive level's reasoning), then counts any remainder:
+
+```text
+NO CODING APPLIED.
+No code reached the 10% confidence threshold, so this record needs human review.
+
+Shelter > Repairs > Roofing — 4%
+  No mention of roof damage; the feedback concerns rent costs.
+
+Water — 3%
+  No reference to water access.
+
+5 further codes scored below 3%.
+```
+
+These explanations are English only, regardless of the language of the feedback.
 
 ## Hyperlinking feedback records
 
--- a/src/qfa/api/schemas.py
+++ b/src/qfa/api/schemas.py
@@ -846,9 +846,11 @@ class ApiAssignedCode(BaseModel):
     explanation: str = Field(
         description=(
             "Judge explanation combining reasoning from all hierarchy "
-            "levels. When no code cleared the confidence threshold, this "
-            "combines every rejected candidate's explanation instead, "
-            "highest-scoring first."
+            "levels. When no code was applied, this instead begins with "
+            "the line 'NO CODING APPLIED.' followed by the reason — empty "
+            "feedback text, no candidate reaching the confidence threshold "
+            "(listing the closest near misses), or nothing in the "
+            "framework being judged relevant. English only."
         )
     )
 
--- a/src/qfa/services/orchestrator.py
+++ b/src/qfa/services/orchestrator.py
@@ -261,17 +261,85 @@ class _ScoredCode:
             for i, (score, expl) in enumerate(zip(self.scores, self.explanations))
         )
 
+    @property
+    def decisive_explanation(self) -> str:
+        """The judge explanation for the level that decided this candidate.
 
-def _combine_rejected_explanations(rejected: list[_ScoredCode]) -> str:
-    """Join every threshold-rejected candidate's explanation into one string.
+        ``_traverse_coding_level`` stops descending at the first level that
+        falls below the threshold, so for a *rejected* candidate the last
+        accumulated level is both its lowest-scoring one and the reason it
+        was dropped. The levels before it passed and would only add noise
+        to a message whose whole point is "why was nothing applied".
+        """
+        return self.explanations[-1]
 
-    Highest-scoring rejection first, each labelled with its hierarchy path
-    so a reader can tell which candidate an explanation belongs to.
+
+NO_CODING_LEAD = "NO CODING APPLIED."
+"""Literal first line of every explanation returned when no code is applied.
+
+EspoCRM surfaces ``assigned_codes.0.explanation`` verbatim as
+``autoCodingExplanation``, so this line is what a user reads first when a
+record comes back uncoded (#256).
+"""
+
+NO_CODING_EMPTY_CONTENT_EXPLANATION = (
+    f"{NO_CODING_LEAD}\nThe feedback text was empty, so there was nothing to code."
+)
+"""Explanation for a record whose ``content`` is empty (issue #138)."""
+
+NO_CODING_NOTHING_RELEVANT_EXPLANATION = (
+    f"{NO_CODING_LEAD}\nNo code in the framework was judged relevant to this feedback."
+)
+"""Explanation for when the LLM selected nothing at any hierarchy level."""
+
+_MAX_LISTED_REJECTIONS = 3
+"""How many near-miss candidates to spell out before collapsing to a count."""
+
+
+def _as_whole_percentage(confidence: float) -> str:
+    """Render a 0-1 confidence as a whole percentage (``0.04`` -> ``"4%"``).
+
+    Non-technical EspoCRM readers see these numbers directly, and a
+    ``0.04`` next to a "Level 2" label reads as noise where "4%" reads as
+    a judgement.
+    """
+    return f"{round(confidence * 100)}%"
+
+
+def _combine_rejected_explanations(
+    rejected: list[_ScoredCode], threshold: float
+) -> str:
+    """Explain in prose why every candidate was rejected by the threshold.
+
+    Leads with :data:`NO_CODING_LEAD` and a sentence naming the threshold,
+    then lists at most :data:`_MAX_LISTED_REJECTIONS` candidates —
+    highest-scoring (closest to being applied) first — as a
+    ``path — percentage`` header over the decisive level's explanation.
+    Any remainder collapses into a single count line rather than an
+    unbounded wall of text.
     """
     ordered = sorted(rejected, key=lambda c: c.confidence_aggregate, reverse=True)
-    return "\n\n".join(
-        f"{' > '.join(name for _, name in c.path)}:\n{c.explanation}" for c in ordered
-    )
+    listed = ordered[:_MAX_LISTED_REJECTIONS]
+
+    blocks = [
+        f"{NO_CODING_LEAD}\n"
+        f"No code reached the {_as_whole_percentage(threshold)} confidence "
+        f"threshold, so this record needs human review."
+    ]
+    blocks += [
+        f"{' > '.join(name for _, name in c.path)} — "
+        f"{_as_whole_percentage(c.confidence_aggregate)}\n"
+        f"  {c.decisive_explanation}"
+        for c in listed
+    ]
+
+    remainder = len(ordered) - len(listed)
+    if remainder:
+        noun = "code" if remainder == 1 else "codes"
+        cutoff = _as_whole_percentage(listed[-1].confidence_aggregate)
+        blocks.append(f"{remainder} further {noun} scored below {cutoff}.")
+
+    return "\n\n".join(blocks)
 
 
 @dataclass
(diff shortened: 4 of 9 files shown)


mariushelf:
`/loop review` received — re-reviewing the PR head as it stands. No fixer round runs; post `/loop fix` if you want one.


mariushelf (commented):
## Review: always explain why no code was applied (#256)

Verified against the ticket's spec and all ten acceptance criteria: **all ten are met**. `make test` (559 passed) and `make lint` (3 import-linter contracts kept) both reproduce green on this head. No blockers, no majors — the findings below are all non-blocking.

### What I checked beyond the diff

- **The `elif rejected and request.confidence_threshold is not None:` narrowing is genuinely dead-code-free.** `_traverse_coding_level` only appends to `rejected` inside `if threshold is not None and judge.score < threshold:` (`orchestrator.py:1581`), so the extra condition can never divert a real below-threshold case into the "nothing relevant" branch. Preferring this over `assert threshold is not None` is the right call in a module with no `assert` precedent.
- **`decisive_explanation = explanations[-1]` is safe and correct.** A rejected candidate is appended at the first sub-threshold level and the loop `continue`s, so every earlier score is `>= threshold` and the last one is both the minimum (matching `confidence_aggregate = min(scores)`) and the decisive one. `explanations` is never empty on that path.
- **`top` empty implies `candidates` empty.** `top = candidates[:max_codes]` could in principle drop real candidates into the no-code branches, which would make "No code in the framework was judged relevant" an actively false statement. `ApiAssignCodesRequest.max_codes` is `ge=1`, so that cannot happen through the API.
- **The known EspoCRM consumer needs no change.** `4_1_assign_codes.php` reads `assigned_codes.0.*` and tolerates nulls, so cases 1 and 3 simply start populating `autoCodingExplanation` where they previously sent nothing.
- **The exact-string test is the right shape for this contract.** `test_message_matches_the_documented_layout` pins the blank-line separators, two-space indent and em-dash headers that substring assertions would let drift.

### Findings

4 minor / nit, ordered by leverage. None gate a merge; they are recorded for the human reviewer.

1. **Minor** — `docs/rest-api/index.md:69-79`: the rendered example is unreachable output.
2. **Minor** — `orchestrator.py:306,339`: `round()` lets a near-miss render *at* the threshold, contradicting the lead line.
3. **Minor** — `schemas.py` / `domain/models.py`: the sibling `coding_level_*` / `confidence_*` descriptions still name the threshold as the only cause of nulls.
4. **Nit** — `.gitignore:203`: `.graphify-out` does not match the `graphify-out/` directory it is meant to ignore.

### Verdict

**Approve.** The change does what the ticket asked, the contract is pinned by tests at both the formatter and behaviour level, and the docs and OpenAPI surfaces were updated in the same PR. Fix (1) whenever convenient — it is the one finding an integrator would actually trip over.

~Reviewed by Claude, run via the agentic engineering loop

docs/rest-api/index.md
```diff
@@ -59,9 +59,26 @@ the single-pass response too.
...
+NO CODING APPLIED.
+No code reached the 10% confidence threshold, so this record needs human review.
+
+Shelter > Repairs > Roofing — 4%
+  No mention of roof damage; the feedback concerns rent costs.
+
+Water — 3%
+  No reference to water access.
+
+5 further codes scored below 3%.
```
mariushelf, line 78: **Minor:** this example cannot be produced by the code it documents.

It lists **two** near misses and then a remainder line for five more — but `_MAX_LISTED_REJECTIONS` is 3, so seven rejections always render three entries. The same paragraph one line above even states "at most three". Rendering the equivalent input through `_combine_rejected_explanations` gives:

```text
Shelter — 4%
  ...

Water — 3%
  ...

X0 — 1%
  ...

4 further codes scored below 1%.
```

Note the count and the cutoff both differ (`4 ... below 1%`, not `5 ... below 3%`), because the cutoff is taken from the *lowest listed* candidate. The PR description and `test_message_matches_the_documented_layout` both carry the correct three-item version — copying that block here would make the doc match the pinned contract exactly. As it stands an integrator sizing their UI off this example expects a shape the API never sends.

src/qfa/services/orchestrator.py
```diff
@@ -261,17 +261,85 @@ def explanation(self) -> str:
...
+
+
+def _as_whole_percentage(confidence: float) -> str:
+    """Render a 0-1 confidence as a whole percentage (``0.04`` -> ``"4%"``).
+
+    Non-technical EspoCRM readers see these numbers directly, and a
+    ``0.04`` next to a "Level 2" label reads as noise where "4%" reads as
+    a judgement.
+    """
+    return f"{round(confidence * 100)}%"
```
mariushelf, line 306: **Minor:** `round()` half-up can make a near miss render *at* the threshold, so the message contradicts its own lead line. With EspoCRM's real `confidence_threshold: 0.1`, a candidate scoring `0.0999`:

```text
NO CODING APPLIED.
No code reached the 10% confidence threshold, so this record needs human review.

A — 10%
  near
```

A non-technical reader now sees a code that reached the threshold in a list of codes that didn't — in a message whose whole purpose is explaining the outcome to that reader. `math.floor(confidence * 100)` for the *candidate* percentages fixes it (keep `round`, or floor too, for the threshold itself) and never over-reports a score.

Two smaller cases from the same source, worth deciding on together:
- Sub-percent scores render `0%` (`0.0004` → `"A — 0%"`), and a threshold below `0.005` renders the lead as "No code reached the **0%** confidence threshold".
- The remainder line at `orchestrator.py:339` inherits the rounding *and* has a ties problem: when the third listed candidate and the unlisted ones share a score, "N further codes scored below 1%" is false — they scored 1%. `at or below` (or deriving the cutoff from the highest *unlisted* score) would be accurate in both cases.

src/qfa/api/schemas.py
```diff
@@ -846,9 +846,11 @@ class ApiAssignedCode(BaseModel):
     explanation: str = Field(
         description=(
             "Judge explanation combining reasoning from all hierarchy "
-            "levels. When no code cleared the confidence threshold, this "
-            "combines every rejected candidate's explanation instead, "
-            "highest-scoring first."
+            "levels. When no code was applied, this instead begins with "
+            "the line 'NO CODING APPLIED.' followed by the reason — empty "
```
mariushelf, line 850: **Minor:** `explanation` was updated for the three-cause contract, but its siblings in this same model were not. `coding_level_1_id`, `coding_level_1_name`, `confidence_level_1` and `confidence_aggregate` (`schemas.py:794-845`) all still read "null when no code cleared the confidence threshold" — which is now only one of three reasons they can be null. These descriptions are OpenAPI-visible, so a client reading the generated schema concludes nulls imply a threshold rejection.

`AssignedCodeModel` in `qfa.domain.models` has the same drift, including a `explanation` description that still says "combines every rejected candidate's explanation instead, highest-scoring first" — the pre-#256 wording, now describing neither the format nor the other two cases. Worth a sweep of both models while the change is fresh, per the repo's "keep docs in sync in the same PR" rule.

.gitignore
```diff
@@ -200,3 +200,5 @@ config
 # See https://github.com/mariushelf/copier-devcontainer
 /.devcontainer/
 /.envrc
+.graphify-out
```
mariushelf: **Nit:** this pattern does not match the directory it is meant to ignore. The artifact directory in the worktree is `graphify-out/` (no leading dot), so it stays untracked-and-noisy despite the `chore: gitignore graphify artifacts` commit. Drop the dot.

Also FYI, unrelated tooling artifacts in a feature PR are the kind of thing that makes a revert less clean — no action needed here, it is two lines.

mariushelf:
Review iteration 1: **approve** — All ten acceptance criteria are met and the change does what the ticket asked. All three no-code paths now return exactly one `assigned_codes` entry with null coding/confidence fields and an explanation leading with the literal `NO CODING APPLIED.`; the below-threshold message is reformatted with whole percentages, the decisive level's reasoning only, at most three near misses sorted highest-first, and a remainder count line omitted when nothing was truncated. I verified the two non-obvious invariants the implementation rests on: `rejected` is only ever appended under `threshold is not None`, so the added `and request.confidence_threshold is not None` guard is a pure type narrowing that cannot divert a real case into the fallback branch; and `explanations[-1]` really is the decisive level because the traversal `continue`s at the first sub-threshold score, making the last level simultaneously the minimum and the cause of rejection. `max_codes` is `ge=1` at the API boundary, so `top` empty implies `candidates` empty — the \"nothing relevant\" message can never fire while real candidates exist. `make test` (559 passed) and `make lint` (3 import-linter contracts kept) both reproduce green on head cce8eda, matching the PR's verification claims. The four findings are all non-blocking: an unreachable rendered example in the integrator-facing REST docs, a `round()`-induced display contradiction for candidates within rounding distance of the threshold, stale sibling field descriptions in the API and domain models, and a `.gitignore` pattern that misses the directory it targets.


mariushelf:
Could not verify: Whether the reformatted message reads well to the actual non-technical EspoCRM audience (wording, tone, the "so this record needs human review" call to action) — that is a product/UX judgment, not something the diff can settle.; Whether any EspoCRM automation or other client beyond `scripts/espo_crm/feedback_trigger/4_1_assign_codes.php` treats `assigned_codes == []` as its "no code" discriminator. The in-repo consumer reads `assigned_codes.0.*` and needs no change, but external or unversioned consumers cannot be checked from here.; Whether English-only explanations are acceptable in production for non-English feedback. The spec explicitly scopes localization out, so this is deliberate, but the user-facing consequence needs a human call.


olafdegraaff:
I think it works, but let me check if this response is not breaking anything in espoCRM first.

olafdegraaff:
Works perfectly LGTM

olafdegraaff (approved):

Issue #267, opened 2026-08-07
refactor(api): delete Orchestrator, rewire composition, and refresh docs


mariushelf:
Part of the `Orchestrator` decomposition epic (#112). Final child — removes the now-empty
shell and brings every live document back in line with the code.

Depends-on: #263
Depends-on: #264
Depends-on: #265
Depends-on: #266

## Spec

**What:** Delete the `Orchestrator` class and its module, remove the
`get_orchestrator` / `build_orchestrator` seam in favour of per-service construction, and
sweep the documentation so nothing still describes the old shape.

**Why:** After the four extractions land, `Orchestrator` is an empty shell that no route
depends on — but `build_orchestrator`, `get_orchestrator`, and roughly two dozen
documents still refer to it. Leaving it in place would preserve exactly the "layer of
indirection readers must traverse for no payoff" the epic set out to remove, and would
leave the docs actively lying about the architecture.

The docs half is **not optional cleanup — it is load-bearing**. `docs/architecture/01-architecture-style.md`
and `docs/architecture/03-components.md` carry hard Sphinx cross-references
(`{py:class}~qfa.services.orchestrator.Orchestrator`,
`{py:func}qfa.api.composition.build_orchestrator`) that will **fail the `make docs`
build** the moment the class is gone.

Two of those pages are *already* stale and should be corrected while they are open:
`01-architecture-style.md` says the class has "four operations" and
`03-components.md` says "five async methods" — it has had six since
`detect_sensitive_content` and `analyze_hierarchical` were added.

### Which documents change, and which must not

- **Update (live docs describing current state):** `AGENTS.md`, `docs/README.md`,
  `docs/architecture/index.md`, `docs/architecture/01-architecture-style.md`,
  `03-components.md`, `04-crosscutting.md`, `05-data-model.md`, `06-prompt-envelope.md`,
  `07-hierarchical-analysis.md`, `docs/development/index.md`,
  `docs/development/implementing-a-new-endpoint.md`,
  `docs/operations/observability.md`, `docs/operations/settings-reference.md`.
- **Do NOT rewrite (point-in-time records):** every file under `docs/adr/` other than the
  ADR-017 pointer already added by the ADR child, and the dated reviews
  `docs/architecture-review-2026-05-08.md` and `-addendum.md`. These are historical
  records of what was decided and observed *then*; editing them to match today's code
  destroys their value. They may keep referring to `Orchestrator` in the past tense.

## Acceptance criteria

[...]

Issue #263, opened 2026-08-07
refactor(services): extract SensitivityService from Orchestrator


mariushelf:
Part of the `Orchestrator` decomposition epic (#112). This is the smallest extraction and
lands first among the four, as the end-to-end proof of the pattern.

Depends-on: #262

## Spec

**What:** Extract `Orchestrator.detect_sensitive_content` (~50 lines, currently around
line 1385) into `SensitivityService` in `src/qfa/services/sensitivity.py`, add a
`get_sensitivity_service` DI provider, and repoint the sensitivity route
(`src/qfa/api/routes.py`, ~line 512) at it.

**Why:** `detect_sensitive_content` is the smallest and most self-contained of the six use
cases — one LLM call, no helper methods, no embedder. That makes it the cheapest possible
first full traversal of the pattern: new service class → DI provider → route rewire →
tests moved → composition updated. Any friction in the approach (import-linter
complaints, DI wiring awkwardness, test-fixture churn) surfaces here on a 50-line change
rather than on the 700-line `AnalyzeService`. The three larger extractions then copy a
proven shape.

## Acceptance criteria

- [ ] `src/qfa/services/sensitivity.py` defines `class SensitivityService:` with **no
      base class**, taking an `LLMCallExecutor` as a constructor dependency.
- [ ] `detect_sensitive_content` is moved onto it and **deleted** from `Orchestrator`.
- [ ] `get_sensitivity_service` is added to `src/qfa/api/dependencies.py`.
- [ ] The sensitivity route handler type-annotates against `SensitivityService` and no
      longer takes `Orchestrator`.
- [ ] `src/qfa/api/composition.py` constructs the service (sharing the one
      `LLMCallExecutor` instance with the still-present `Orchestrator`).
- [ ] Tests for this use case move to `tests/services/test_sensitivity.py`, constructing
      the real `LLMCallExecutor` over `FakeLLMPort` / `FakeAnonymizer`. No test assertion
      is weakened or deleted in the move.
- [ ] The existing e2e/API tests for the sensitivity endpoint pass **unmodified** —
      response shape, status codes, and error payloads are unchanged.
- [ ] `make test` and `make lint` pass, including all three `import-linter` contracts.

mariushelf:
~Written by Claude, run via the agentic engineering loop

mariushelf:
**Escalated:** ticket qfa#263 crashed with an uncaught exception.

```
[...]

PR #271, opened 2026-08-10
refactor(services): extract LLMCallExecutor collaborator from Orchestrator

### LINKED ISSUE #262: refactor(services): extract LLMCallExecutor collaborator from Orchestrator
Part of the `Orchestrator` decomposition epic (#112). This is the shared foundation every
service extraction builds on, so it lands second (right after the ADR) and changes no
public behaviour.

Depends-on: #261

## Spec

**What:** Extract the LLM-call scaffolding currently spread across `Orchestrator`'s
private methods into a new plain class `LLMCallExecutor` in
`src/qfa/services/llm_call_executor.py`. `Orchestrator` keeps all six public methods with
identical signatures and delegates the extracted work to an injected executor.

Methods that move:

| Method | Current line | Role |
|---|---|---|
| `_anonymize_records` | ~839 | PII redaction before the LLM call |
| `_bounded_complete` | ~860 | the retrying, deadline-bounded LLM call |
| `_check_deadline_and_get_timeout` | ~1435 | per-call timeout derivation |
| `_check_token_limit` | ~1600 | pre-flight token budget guard |

**Why:** Every use-case service needs this same scaffolding. Deciding *how* it is shared
is the whole architectural question this epic answers, and #112 settled it: a single
injected collaborator, not a base class. Landing the collaborator on its own — while
`Orchestrator` is still intact and its full test suite still green — proves the seam is
correct before four extractions commit to it. If the executor's boundary is wrong, this
is the cheapest possible place to discover it.

No route, DI, or response change happens here.

## Acceptance criteria

- [ ] `src/qfa/services/llm_call_executor.py` defines `class LLMCallExecutor:` with
      **no base class** and no `Protocol`.
- [ ] `LLMCallExecutor` is **not** referenced from `qfa/domain/ports.py`.
- [ ] Its constructor takes the driven ports and config it needs — `llm: LLMPort`,
      `anonymizer: AnonymizationPort`, `settings: OrchestratorSettings`,
      `llm_timeout_seconds: float`, `max_total_tokens: int`.
- [ ] The four methods above are moved onto it (renamed to public names where they are
      now cross-class calls) and **deleted** from `Orchestrator`.
- [ ] `Orchestrator` holds an `LLMCallExecutor` and delegates; all six public methods
      (`analyze_bulk`, `analyze_hierarchical`, `summarize_bulk`, `summarize`,
      `assign_codes`, `detect_sensitive_content`) keep byte-identical signatures.
- [ ] `build_orchestrator` in `src/qfa/api/composition.py` constructs the executor and
      passes it in; `build_orchestrator`'s own signature is unchanged.
- [ ] New `tests/services/test_llm_call_executor.py` covers the executor directly:
      retry-on-transient, non-transient passthrough, deadline expiry, token-limit
      rejection, and anonymize→deanonymize round-tripping — constructed over the existing
      `FakeLLMPort` / `FakeAnonymizer`, not over a new fake executor.
- [ ] `tests/services/test_orchestrator.py` and
      `tests/services/test_orchestrator_hierarchical.py` still pass with changes limited
      to how the `Orchestrator` under test is constructed. No test assertion is weakened
      or deleted to accommodate the refactor.
- [ ] Behaviour-preserving: no HTTP request/response shape, status code, or error payload
      changes.

mariushelf:
Tree is clean, one commit on `feat/262-loop-refactor-services-extract-llmcallexecuto`.

## PR-ready summary

**refactor(services): extract LLMCallExecutor collaborator from Orchestrator** — closes #262

### What changed

New `src/qfa/services/llm_call_executor.py` holds `class LLMCallExecutor:` (no base class, no `Protocol`, not referenced from `qfa/domain/ports.py`) with the four LLM-call helpers that were spread across `Orchestrator`'s privates, renamed to public names now that the calls are cross-class:

| Was on `Orchestrator` | Now on `LLMCallExecutor` |
|---|---|
| `_anonymize_records` | `anonymize_records` |
| `_bounded_complete` | `bounded_complete` |
| `_check_deadline_and_get_timeout` | `check_deadline_and_get_timeout` |
| `_check_token_limit` | `check_token_limit` |

Constructor takes `llm`, `anonymizer`, `settings`, `llm_timeout_seconds`, `max_total_tokens`. `_SlotTiming` moved with `bounded_complete` and became public `SlotTiming` — it is now a cross-module signature type. `_MINIMUM_ATTEMPT_WINDOW` moved and stayed private to the new module.

`Orchestrator` holds `self._executor` and delegates at all 13 call sites; the four methods are deleted from it (1734 → 1617 lines). `build_orchestrator` constructs the executor and passes it in; its own signature is unchanged.

Two judgement calls worth a reviewer's eye:

- **`executor` is an optional constructor argument**, default-constructed from the ports `Orchestrator` already receives. There are ~60 `Orchestrator(...)` sites across tests, `scripts/`, and notebooks; making it required would churn all of them for no behavioural gain, and the spec's AC bounds test changes to how the orchestrator under test is constructed. The composition root injects explicitly, per ADR-017; everyone else gets an equivalent one for free. `tests/services/test_orchestrator.py` and `test_orchestrator_hierarchical.py` needed **zero** changes as a result.
- **`bounded_complete`'s `llm` argument is now optional**, defaulting to the connection the executor was built over. The spec requires `llm` in the constructor, but none of the four moved methods used `self._llm` — this makes it load-bearing rather than vestigial. Every existing call site still passes `llm=` explicitly, so judge calls stay visibly routed to the judge connection and behaviour is identical.

### Tests

New `tests/services/test_llm_call_executor.py` (21 tests) drives the **real** executor over the existing `FakeLLMPort` / `FakeAnonymizer` doubles imported from `test_orchestrator` — no fake executor. Covers deadline expiry and the insufficient-window guard, retry-budget reservation (`remaining / LLM_RETRY_BUDGET_MULTIPLIER`), transient errors (`LLMTimeoutError`/`LLMRateLimitError`) propagating with exactly one attempt, non-transient `LLMError` passthrough, token-limit rejection with `estimated_tokens`/`limit` plus the at-limit boundary, anonymize→deanonymize round-tripping, the primary/override client split, semaphore bounding, and timing capture (including the deadline-expires-during-queue-wait case, which pins that the timeout is derived *after* slot acquisition).

One test double was added: `RedactingAnonymizer` — the shared `FakeAnonymizer` is a deliberate no-op, so it cannot demonstrate a round-trip. It inherits `AnonymizationPort` per AGENTS.md.

### Docs

`docs/architecture/03-components.md` gains a "The LLM-call executor" subsection (method table, why it's neither port nor base class, primary-vs-judge connection) and the composition-root bullet now names the executor. `docs/architecture/04-crosscutting.md`: the deadline-check row now attributes the concern to the executor, plus a note on batch vs single-message redaction. `docs/development/implementing-a-new-endpoint.md`: the worked example calls `self._executor.check_deadline_and_get_timeout(...)`, with a paragraph telling authors to reuse the executor rather than re-derive the arithmetic. `qfa.settings.LLM_RETRY_BUDGET_MULTIPLIER`'s comment now names the executor as the constant's second reader.

### Impact / risk

Low. Behaviour-preserving refactor: all six public method signatures verified byte-identical against `HEAD`, no HTTP request/response shape, status code, or error payload change, no route or DI change. `make test` — 613 passed. `make lint` — ruff, `ty`, and all three import-linter contracts pass. `make docs` builds clean under `-W`.

The main residual risk is the coexistence window ADR-017 anticipates: `Orchestrator` still calls `self._llm` / `self._anonymizer` directly in the paths that were not part of the four moved helpers, so the executor is not yet the sole route to the LLM. That converges in #263–#267. A caller that injects an executor built over a *different* LLM than the orchestrator's would get divergent routing — the composition root doesn't, and the docstring flags it.

Fixes #262

### DIFF
--- a/src/qfa/services/llm_call_executor.py
+++ b/src/qfa/services/llm_call_executor.py
@@ -1,0 +1,216 @@
+"""Shared LLM-call scaffolding for the application services.
+
+Every use case in :mod:`qfa.services` wraps its LLM calls in the same
+four concerns: redact PII before the call, derive a per-call timeout from
+the request deadline, guard the token budget, and bound concurrent calls
+with a semaphore. :class:`LLMCallExecutor` owns those four and nothing
+else, so a use-case service can be read without them.
+
+Per ADR-017 this is a plain concrete class, deliberately **not** a
+Protocol and **not** a base class:
+
+- It is not a driven port. Ports in this codebase invert dependencies on
+  *infrastructure*; this object wraps no external system, it orchestrates
+  calls to an :class:`~qfa.domain.ports.LLMPort` that already exists. It
+  is therefore not declared in :mod:`qfa.domain.ports`.
+- Services receive it as a constructor dependency and delegate to it.
+  Nothing inherits from it — behaviour reuse in this codebase is always
+  composition, and ``class X(Y):`` stays readable as "X implements port
+  Y".
+
+Tests construct the *real* executor over the existing ``FakeLLMPort`` /
+``FakeAnonymizer`` doubles; there is no fake executor.
+"""
+
+import asyncio
+import time
+from dataclasses import dataclass
+from datetime import UTC, datetime
+
+from qfa.domain.errors import AnalysisTimeoutError, FeedbackTooLargeError
+from qfa.domain.models import FeedbackRecordModel, LLMResponse, T_Response
+from qfa.domain.ports import AnonymizationPort, LLMPort
+from qfa.settings import LLM_RETRY_BUDGET_MULTIPLIER, OrchestratorSettings
+
+#: Minimum time (seconds) required for an LLM attempt to be viable.
+_MINIMUM_ATTEMPT_WINDOW = 10.0
+
+
+@dataclass
+class SlotTiming:
+    """Split timing for one semaphore-bounded hierarchical LLM call.
+
+    ``queued_seconds`` is the time spent waiting to acquire the concurrency
+    semaphore; ``call_seconds`` is the LLM completion itself, measured only
+    *after* the slot was acquired. Keeping them apart makes the per-chunk
+    debug lines honest: a single combined duration folds queue-wait into
+    "call time", so a 5s call that waited 100s in the queue looked like a
+    110s call (and like a near-timeout when it was nothing of the sort).
+    """
+
+    queued_seconds: float = 0.0
+    call_seconds: float = 0.0
+
+
+class LLMCallExecutor:
+    """Deadline-, budget- and concurrency-aware wrapper around an LLM port.
+
+    Parameters
+    ----------
+    llm : LLMPort
+        The LLM provider adapter every call runs on by default. Call sites
+        that must use a *different* connection — the LLM-as-judge calls,
+        which may be configured onto their own model — pass it explicitly
+        to :meth:`bounded_complete`.
+    anonymizer : AnonymizationPort
+        The anonymisation adapter used to redact PII before LLM calls.
+    settings : OrchestratorSettings
+        Cross-cutting configuration; this object reads ``chars_per_token``
+        for its token estimate.
+    llm_timeout_seconds : float
+        Maximum time in seconds for a single LLM call, before the deadline
+        is taken into account.
+    max_total_tokens : int
+        Maximum estimated total tokens for a single request.
+    """
+
+    def __init__(
+        self,
+        llm: LLMPort,
+        anonymizer: AnonymizationPort,
+        settings: OrchestratorSettings,
+        llm_timeout_seconds: float,
+        max_total_tokens: int,
+    ) -> None:
+        self._llm = llm
+        self._anonymizer: AnonymizationPort = anonymizer
+        self._settings = settings
+        self._llm_timeout_seconds = llm_timeout_seconds
+        self._max_total_tokens = max_total_tokens
+
+    def anonymize_records(
+        self,
+        records: tuple[FeedbackRecordModel, ...],
+        anonymize: bool,
+    ) -> tuple[tuple[FeedbackRecordModel, ...], dict[str, str]]:
+        """Anonymise each record's text, returning new records + merged mapping.
+
+        Metadata is left untouched (codes/dates are not PII and feed the
+        deterministic trend table). When ``anonymize`` is False, records are
+        returned unchanged with an empty mapping.
+        """
+        if not anonymize:
+            return records, {}
+        merged: dict[str, str] = {}
+        new_records: list[FeedbackRecordModel] = []
+        for record in records:
+            redacted, mapping = self._anonymizer.anonymize(record.content)
+            merged.update(mapping)
+            new_records.append(record.model_copy(update={"content": redacted}))
+        return tuple(new_records), merged
+
+    async def bounded_complete(
+        self,
+        semaphore: asyncio.Semaphore,
+        *,
+        llm: LLMPort | None = None,
+        system_message: str,
+        user_message: str,
+        tenant_id: str,
+        response_model: type[T_Response],
+        deadline: datetime,
+        timing: SlotTiming | None = None,
+    ) -> LLMResponse[T_Response]:
+        """Run one LLM completion, bounded by ``semaphore`` and the deadline.
+
+        ``semaphore`` caps how many completions run at once across the whole
+        hierarchical pipeline (map, leaf judge, reduce), so concurrency stays
+        within ``max_concurrent_chunks`` across every phase. The
+        deadline/timeout is computed *after* acquiring a slot, so a
+        completion that queued behind others still honours the remaining budget
+        (and raises ``AnalysisTimeoutError`` if the deadline passed while it
+        waited).
+
+        ``llm`` overrides the connection for this one call. It exists because
+        this helper serves both map/reduce and the leaf judges, which may run
+        on different connections (see ``judge_llm``); ``None`` uses the
+        executor's own client. Every current call site passes it explicitly so
+        the connection each call uses is visible where the call is written.
+        Note the semaphore is shared regardless: the bound is on total
+        in-flight calls, not per connection.
+
+        When ``timing`` is supplied it is populated with the queue-wait and the
+        post-acquire call duration as two separate fields, so callers can log
+        them apart rather than reporting one combined number that hides how long
+        the call sat waiting for a slot.
+        """
+        client = llm if llm is not None else self._llm
+        queue_start = time.perf_counter()
+        async with semaphore:
+            acquired_at = time.perf_counter()
+            if timing is not None:
+                timing.queued_seconds = acquired_at - queue_start
+            # Compute the timeout only now: queue-wait already elapsed, so the
+            # per-call window reflects the budget that actually remains.
+            timeout = self.check_deadline_and_get_timeout(deadline)
+            try:
+                return await client.complete(
+                    system_message=system_message,
+                    user_message=user_message,
+                    tenant_id=tenant_id,
+                    response_model=response_model,
+                    timeout=timeout,
+                )
+            finally:
+                if timing is not None:
+                    timing.call_seconds = time.perf_counter() - acquired_at
+
+    def check_deadline_and_get_timeout(self, deadline: datetime) -> float:
+        """Raise if the deadline has passed or too little time remains.
+
+        Return a timeout (seconds) bounded by the deadline and the
+        configured per-call limit.
+        """
+        remaining = (deadline - datetime.now(UTC)).total_seconds()
+        if remaining <= 0:
+            raise AnalysisTimeoutError("Deadline exceeded")
+        if remaining < _MINIMUM_ATTEMPT_WINDOW:
+            raise AnalysisTimeoutError(
+                f"Insufficient time remaining ({remaining:.1f}s) for an LLM attempt"
+            )
+        # A single ``LLMPort.complete`` may retry internally, spending up to
+        # ``LLM_RETRY_BUDGET_MULTIPLIER`` times this per-attempt timeout in worst
+        # case (see ``qfa.adapters.llm_client``). Divide the remaining deadline
+        # by the same factor so even a fully-retried last call finishes before
+        # the deadline. With a generous deadline the per-call cap binds first, so
+        # this only bites as the deadline approaches.
+        per_attempt_budget = remaining / LLM_RETRY_BUDGET_MULTIPLIER
+        return min(self._llm_timeout_seconds, per_attempt_budget)
+
+    def check_token_limit(self, system_message: str, user_message: str) -> None:
+        """Estimate total tokens and raise if over the limit.
+
+        Parameters
+        ----------
+        system_message : str
+            The assembled system message.
+        user_message : str
+            The assembled user message containing the feedback records.
+
+        Raises
+        ------
+        FeedbackTooLargeError
+            When estimated tokens exceed the configured limit.
+        """
+        assembled_text = system_message + user_message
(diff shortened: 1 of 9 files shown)


mariushelf:
Build complete. 

mariushelf:
Builder committed 1 of the 6 tasks its own plan (`tasks/plan.md`) declares. This is a signal for the review round, not a verdict: a short count *can* mean the build was cut off and the remaining tasks were never built, but a verification-only task commits nothing and one commit can cover several tasks. Check the diff against the spec's acceptance criteria before trusting the summary above.

src/qfa/services/orchestrator.py
```diff
@@ -440,6 +439,18 @@ def __init__(
         self._analyze_settings = analyze_settings or AnalyzeSettings()
         self._llm_timeout_seconds = llm_timeout_seconds
```
mariushelf, line 440: **minor — dead state, and it makes an existing regression guard hollow.**

After this refactor nothing reads `Orchestrator._llm_timeout_seconds`: the only consumer was `_check_deadline_and_get_timeout`, which now lives on the executor and uses the executor's own copy. (`self._max_total_tokens` and `self._settings` are still genuinely read — lines 680, 1059, 1136 — so this is specific to the timeout.)

The consequence isn't just an unused attribute. `tests/api/test_composition.py:371` guards the factory wiring with:

```python
assert orchestrator._llm_timeout_seconds == expected_timeout
```

That assertion now passes regardless of what `build_orchestrator` wires into the `LLMCallExecutor` — the value it checks no longer influences any LLM call. A regression that passed the wrong `llm_timeout_seconds` to the executor (the value that actually caps every call) would go undetected, and the test's own docstring says "a regression here would silently cap the wrong chunk size — guard it."

Suggested: drop `self._llm_timeout_seconds` from `Orchestrator` and point the composition test at `orchestrator._executor._llm_timeout_seconds`. Flagging rather than blocking — nothing is broken today, but the guard is now decorative.

src/qfa/services/llm_call_executor.py
```diff
@@ -0,0 +1,216 @@
...
+    def __init__(
+        self,
+        llm: LLMPort,
+        anonymizer: AnonymizationPort,
+        settings: OrchestratorSettings,
+        llm_timeout_seconds: float,
+        max_total_tokens: int,
+    ) -> None:
+        self._llm = llm
+        self._anonymizer: AnonymizationPort = anonymizer
```
mariushelf, line 86: **minor — two sources of truth for the same collaborators during the coexistence window.**

The executor holds its own `_anonymizer`, `_settings` and `_max_total_tokens`, while `Orchestrator` keeps reading `self._anonymizer` / `self._settings` / `self._max_total_tokens` directly for the paths that weren't part of the four moved helpers. `analyze_hierarchical` now round-trips across both: it redacts via `self._executor.anonymize_records(...)` (orchestrator.py:637) and restores via `self._anonymizer.deanonymize(...)` (orchestrator.py:904).

With an injected executor built over a different anonymizer, that round-trip silently fails to restore — placeholders would leak into the response body. `build_orchestrator` passes the same `PresidioAnonymizer` instance to both, and the default-constructed executor reuses the orchestrator's arguments, so no live path can diverge today; the PR body and the class docstring both flag it, and #263–#267 converge it.

Not asking for a change here — the epic's sequencing owns this. Worth a human's eye because there is no invariant (assertion, or deriving one from the other) preventing an inconsistent pair from being constructed, and the failure mode is a PII-shaped one.

src/qfa/services/llm_call_executor.py
```diff
@@ -0,0 +1,216 @@
...
+        anonymize: bool,
+    ) -> tuple[tuple[FeedbackRecordModel, ...], dict[str, str]]:
+        """Anonymise each record's text, returning new records + merged mapping.
+
+        Metadata is left untouched (codes/dates are not PII and feed the
+        deterministic trend table). When ``anonymize`` is False, records are
+        returned unchanged with an empty mapping.
+        """
+        if not anonymize:
+            return records, {}
```
mariushelf, line 103: **nit — scope: `llm` becoming optional wasn't asked for.**

The ticket moves `_bounded_complete` as-is; making its `llm` argument optional is an added affordance. The rationale in the PR body is sound (the constructor `llm` would otherwise be vestigial, and the AC does require it), all three call sites still pass `llm=` explicitly, and `Orchestrator._judge_llm` is never `None` (it falls back to `llm` at orchestrator.py:432), so the implicit-fallback branch is unreachable from production code.

The cost is that a future call site can now omit `llm=` and silently get the primary connection where a judge connection was intended — exactly what the old required argument was there to prevent ("It stays a required argument so each call site states which client it uses"). The new docstring keeps the convention as prose rather than as a signature. Fine as a judgement call; recording it so it's a deliberate one.

tests/services/test_llm_call_executor.py
```diff
@@ -0,0 +1,393 @@
...
+
+    @pytest.mark.asyncio
+    @pytest.mark.parametrize(
+        "error",
+        [
+            LLMTimeoutError("upstream timed out"),
+            LLMRateLimitError("429 from provider"),
+        ],
+    )
+    async def test_transient_errors_propagate_without_a_second_attempt(self, error):
```
mariushelf, line 311: **nit — this is the AC's "retry-on-transient", inverted.**

The acceptance criteria ask the new suite to cover "retry-on-transient", and the ticket's table describes `_bounded_complete` as "the retrying, deadline-bounded LLM call". Neither is quite true of the code: retry lives in `LiteLLMClient` via `tenacity`, and the moved method never retried. This test pins the opposite — one attempt, error propagates — which is the correct assertion for this class, and `test_tight_deadline_reserves_the_adapter_retry_budget` covers the executor's actual retry-related responsibility (dividing the deadline budget by `LLM_RETRY_BUDGET_MULTIPLIER`).

So the intent behind the AC is covered; the AC's wording just described the wrong owner. Noting it so the mismatch between checklist and tests reads as deliberate rather than as a gap.

mariushelf (commented):
## Review — PR #271 (issue #262): extract `LLMCallExecutor`

**Verdict: approve.** This does what the ticket asked, and the refactor is genuinely behaviour-preserving as far as I can verify. Findings below are all non-blocking.

### On the loop's "1 of 6 tasks" note

False alarm. `tasks/plan.md` lists 6 *steps*, not 6 commits: step 6 is "`make lint`, `make test`. Commit." — i.e. the plan explicitly ends in a single commit covering steps 1–5. All five substantive steps are present in `0d55b25`: the new module, the `Orchestrator` delegation, the composition-root wiring, the new test file, and all three doc pages. Nothing was cut off.

### What I verified

| Check | Result |
|---|---|
| Six public signatures byte-identical vs `ecf8c9e` | ✅ diffed each one mechanically — no change |
| `build_orchestrator` signature unchanged | ✅ identical |
| `LLMCallExecutor` referenced from `qfa/domain/ports.py` | ✅ no reference |
| Four methods deleted from `Orchestrator` | ✅ no residual definitions or call sites of the old private names |
| `make test` | ✅ 613 passed, 42 deselected |
| `make lint` | ✅ ruff, `ty`, and 3/3 import-linter contracts kept |
| `make docs` | ✅ builds clean; new module picked up by autodoc |
| `test_orchestrator.py` / `test_orchestrator_hierarchical.py` | ✅ **zero** changes — the diff touches no existing test file |

The last row is worth calling out: the AC allowed changes to how the orchestrator under test is constructed, and the build needed none, because `executor` defaults to one built from the arguments `Orchestrator` already receives. That's the strongest available evidence the seam is behaviour-preserving — ~60 construction sites and the entire existing suite exercise the new delegation path unmodified, with no assertion touched.

### Judgement calls I agree with

- **Optional `executor` kwarg.** Purist DI would make it required, but the composition root injects explicitly (satisfying ADR-017 where it matters), and requiring it would have churned every test/script/notebook construction site for no behavioural gain — churn that would also have destroyed the "existing tests unchanged" evidence above.
- **`SlotTiming` made public.** It's now a cross-module signature type; keeping it `_`-prefixed would be worse.
- **`RedactingAnonymizer` as a new double.** Justified — the shared `FakeAnonymizer` is a no-op and literally cannot demonstrate a round-trip. It inherits `AnonymizationPort` per AGENTS.md.

The new suite is good: it pins the semaphore bound, the timing split, and specifically that the timeout is derived *after* slot acquisition (`test_deadline_that_expires_during_queue_wait_is_honoured`) — a property that was previously only implicit in the method body.

### Findings

| # | Severity | Where | What |
|---|---|---|---|
| 1 | minor | `src/qfa/services/orchestrator.py:440` | `Orchestrator._llm_timeout_seconds` is now dead state, and `tests/api/test_composition.py:371` still asserts on it — that regression guard no longer covers the value any LLM call actually uses |
| 2 | minor | `src/qfa/services/llm_call_executor.py` | Executor and `Orchestrator` each hold their own anonymizer/settings/token-ceiling during the coexistence window; nothing enforces they agree, and `analyze_hierarchical` round-trips across both |
| 3 | nit | `src/qfa/services/llm_call_executor.py:103` | `bounded_complete(llm=...)` becoming optional is out of scope and softens a deliberate "state your client" constraint into prose |
| 4 | nit | `tests/services/test_llm_call_executor.py:311` | The AC's "retry-on-transient" is covered inverted (executor must *not* retry) — correct for the code, but the checklist wording described the wrong owner |

Details inline.

### Not verified

- Behaviour preservation is argued from identical moved logic plus a green suite (including `tests/api/`), not from a live HTTP diff against `main`. No request/response model, route, status code, or error payload is touched in the diff, and the two error types the moved code raises (`AnalysisTimeoutError`, `FeedbackTooLargeError`) keep identical messages and fields — but a byte-level response comparison against a running instance is a human call.

mariushelf:
Review iteration 1: **approve** — The PR does what the ticket asked. `LLMCallExecutor` is a plain concrete class (no base class, no Protocol, not referenced from `qfa/domain/ports.py`) with exactly the specified constructor, the four methods are moved onto it and deleted from `Orchestrator`, and delegation happens at all 13 call sites. I mechanically diffed all six public `Orchestrator` method signatures and `build_orchestrator`'s signature against the base commit (`ecf8c9e`) — byte-identical. `make test` (613 passed), `make lint` (ruff, `ty`, 3/3 import-linter contracts kept) and `make docs` all pass locally. Notably the diff touches *no* existing test file: `test_orchestrator.py` and `test_orchestrator_hierarchical.py` needed zero changes because `executor` defaults to one built from the arguments `Orchestrator` already receives, which is the strongest available evidence the seam preserves behaviour. The loop's "1 of 6 tasks committed" note is a false alarm — `tasks/plan.md` lists 6 steps whose sixth is "run lint/test, commit", and all five substantive steps (new module, delegation, composition wiring, new tests, three doc pages) are present in the single commit. Two minor findings and two nits, none of which should block a human reviewer: the refactor leaves `Orchestrator._llm_timeout_seconds` as dead state, which silently hollows out an existing regression guard in `tests/api/test_composition.py:371`; and the coexistence window leaves the executor and the orchestrator each holding their own anonymizer/settings/token-ceiling with no invariant tying them together.


mariushelf:
Could not verify: Behaviour preservation is argued from identical moved logic plus a fully green suite (including tests/api/), not from a live HTTP request/response diff against a running instance on main. No route, response model, status code or error payload is touched in the diff, and the two error types the moved code raises (AnalysisTimeoutError, FeedbackTooLargeError) keep identical messages and fields — but a byte-level response comparison is a human call.; Whether the intentionally-optional `executor` constructor argument is acceptable long-term DI style for this codebase, versus making it required once #263–#267 land. The PR argues it explicitly and the composition root injects; this is a taste/architecture call for a human.


olafdegraaff:
LGTM

PR #273, opened 2026-08-11
feat(coding): assign codes in one shot instead of per-level traversal

olafdegraaff:
## Summary
- Replace the recursive pick-then-judge-per-level coding flow with a single LLM call per feedback record
- The LLM selects the best-fitting code(s) directly from the flattened set of all coding-framework nodes (any depth), self-reporting confidence, instead of drilling down level by level with a separate judge call at each step

## Test plan
- [ ] `make test`
- [ ] `make lint`
- [ ] Manually verify `assign_codes` output against a multi-level coding framework


olafdegraaff:
## Code review

Automated review of the one-shot `assign_codes` rewrite. Findings, most severe first:

### Correctness

1. **`src/qfa/services/orchestrator.py:1439`** — The one-shot coding call lost the old graceful degradation on malformed LLM output. The previous `pick` step (`response_model=str` + `_parse_selected_indices`) swallowed non-JSON/malformed responses and returned `[]` (→ "no codes selected"). The new code validates the whole response as `CodingResponse` via strict `model_validate_json`; any malformed field anywhere in `selected` now raises a `ValidationError`, wrapped as `LLMError`, which surfaces as a 502 for that record instead of a normal 200 with "nothing relevant."

2. **`src/qfa/services/coding_classifier.py:52`** — Out-of-range confidence from the LLM now raises `LLMError` (`code="llm_error"`) instead of the old explicit `AnalysisError` (`code="analysis_unavailable"`). The removed manual `0.0 <= score <= 1.0` check is only replaced by pydantic's `ge=0.0, le=1.0` on the *incoming* side — the outgoing schema has `minimum`/`maximum` stripped (`_provider_safe_response_format`), so the LLM isn't constrained either. Any consumer branching on `error.code` for this case silently stops matching.

3. **`src/qfa/services/orchestrator.py:1474`** — A selected code path deeper than 3 levels is silently truncated to its level-3 ancestor. `flatten_coding_nodes` supports arbitrary depth, but only `path[0..2]` are used to populate `coding_level_1/2/3` — a depth-4 selection gets misreported as its shallower ancestor with no validation or warning.

4. **`src/qfa/services/orchestrator.py:1446`** — Invalid/duplicate selection indices from the LLM are dropped silently. If *every* index is invalid, the response falls through to `NO_CODING_NOTHING_RELEVANT_EXPLANATION`, masking a malformed-response failure as a legitimate "nothing relevant" business outcome.

### Docs drift

5. **`docs/operations/settings-reference.md:25`** — Still describes a "per-level judge inside `assign_codes`" that this PR removes entirely; sibling docs (`03-components.md`, `rest-api/index.md`) were updated in this diff, this one wasn't (per AGENTS.md's docs-sync rule).

6. **`src/qfa/domain/models.py:308`** — `AssignedCodeModel`'s docstring still says "a single leaf code," but the new design intentionally allows non-leaf selections as valid final answers.

### Efficiency

7. **`src/qfa/services/orchestrator.py:1410`** — The one-shot call now sends every node at every depth of the framework as options in a single prompt (vs. one level's labels per call previously). A framework sized to pass every old per-level token check could now trip `FeedbackTooLargeError` on the combined call — undocumented and untested.

### Simplification / cleanup

8. **`coding_classifier.py:24`** — `CodePathOption.label` and `orchestrator.py`'s `_CodeCandidate.label` duplicate the identical `" > ".join(...)` expression across two files.
9. **`orchestrator.py:1470`** — `confidence_level_1/2/3` and `coding_level_2/3_id/name` are populated via six near-identical `len(c.path) > N` ternaries — a small helper would collapse these.
10. **`orchestrator.py:1415`** — The "no options in framework" early return hand-duplicates the `CodedFeedbackRecordModel`/`CodingAssignmentResultModel` construction that the shared tail already builds for the "nothing selected" case.
11. **`coding_classifier.py:30`** — `flatten_coding_nodes`'s recursive accumulator parameter is named `_prefix`, using the leading-underscore "internal, don't pass" convention for a value that's actually passed on every recursive call.
12. **`orchestrator.py:1410`** — `list(request.coding_levels.root_codes)` wraps an already-`list[CodingNode]` value in a redundant copy.

---
*Generated by an automated review pass (Claude Code).*

olafdegraaff (commented):
## Automated review (7-angle pass: diff scan, removed-behavior audit, cross-file trace, reuse, simplification, efficiency, altitude)

**Breaking-change verdict:** No hard API break — response shape is unchanged (`confidence_level_1/2/3`, `confidence_aggregate` fields still all present, no symbol removed/renamed, `LLMResponseParseError` inherits from `LLMError` so every other endpoint's error handling is unaffected). However there's a **silent semantic break**: `confidence_level_2`, `confidence_level_3`, and `confidence_aggregate` now all mirror the single self-reported `confidence_level_1` value instead of being independently judged / `min()`-aggregated as before. Any consumer (dashboards, EspoCRM exports) that used divergence between levels as a signal will silently get three identical numbers with no error. Also, the old design's structural guarantee that a child code could never be selected unless its ancestors first cleared the threshold is gone, replaced by prompt wording only — worth a deliberate sign-off even though nothing crashes.

### High
1. **Ancestor-gate pruning removed, no structural replacement** — `src/qfa/services/coding_classifier.py:29-42`, `src/qfa/services/orchestrator.py:1440-1477`. Every node at every depth is now an independent, atomically-scored option; a level-3 leaf can be selected/thresholded-in even though its level-1/2 ancestors would have failed the old per-level judge. Enforcement is prompt-only (`_SYSTEM` in `coding_classifier.py:67-100`), with no post-hoc consistency check.
2. **No dedup across selections for parent/child (or otherwise overlapping) paths** — `src/qfa/services/orchestrator.py:1455-1477`. `seen_indices` only blocks exact duplicate indices. The LLM can legally return both "Service Delivery" and "Service Delivery > Staff Behavior > Rudeness" for one record; both survive into `assigned_codes`, double-counting one idea downstream (e.g. in `coding_trends.py`). Old recursive design made this structurally impossible.
3. **One out-of-range confidence aborts the whole request** — `src/qfa/services/orchestrator.py:1464-1465`. A single bad float among several otherwise-valid `CodeSelection`s raises `AnalysisError` and discards all selections for that record, not just the offending one.
4. **Schema-validation failures swallowed with zero logging** — `src/qfa/services/orchestrator.py:1449-1453`. Unlike the hierarchical leaf-judge failure path (`logger.warning(...)` at ~line 1058), a persistent `LLMResponseParseError` here silently degrades to `selections = []` → "NO CODING APPLIED... nothing relevant", indistinguishable from a genuine no-match and invisible in logs/metrics.

### Medium
5. **Full framework flattened into one prompt** — `src/qfa/services/orchestrator.py:1415-1436`. Previously only one level's children were ever embedded per call; now every node at every depth goes into a single `user_message`. For large frameworks this can newly trip `_check_token_limit`/`FeedbackTooLargeError` for requests that succeeded fine under the old per-level design.
6. **Anonymizer runs full NER over the entire flattened framework text, not just the feedback text** — `src/qfa/services/orchestrator.py:1437` via `build_coding_messages` (`coding_classifier.py:105-119`). Framework labels are admin config, never PII, but now get scanned by `langdetect`/spaCy on every call — pure wasted inference vs. the old bounded per-level label set.
7. **Stale docstring: `max_codes` description** — `src/qfa/domain/models.py:293-296` still says "Maximum number of leaf codes to retain," but `flatten_coding_nodes` now exposes non-leaf nodes as selectable options too, so the cap can bind on intermediate-level codes.

### Low / doc-rot / cleanup
8. Stale doc: `docs/architecture/03-components.md:62` still says "...the whole of `assign_codes` including its per-level judge stays on the generation client" — there's no judge call left in `assign_codes` at all (one-shot only). Line 107 in the same file was correctly updated.
9. Stale docstring: `src/qfa/services/orchestrator.py:376-379` — "The per-level judge inside `assign_codes` deliberately stays on `llm`" describes a step that no longer exists.
10. Duplicated path-label join logic: `CodePathOption.label` (`coding_classifier.py:23-26`) and `_CodeCandidate.label` (`orchestrator.py:252-254`) both do `" > ".join(name for _, name in path)` — a future formatting change (escaping, truncation, localization) needs updating in two places.
11. Duplicated tree-walk: `flatten_coding_nodes` (`coding_classifier.py:29-42`) is a third independent recursive walker over the same `id/name/children` shape already traversed by `ApiCodingNode.max_child_depth`/`min_child_depth` (`src/qfa/api/schemas.py:755-765`).
12. Confidence-range check (`0.0 <= x <= 1.0`) is re-derived inline at `orchestrator.py:1464` rather than reusing a shared helper — same check existed verbatim in the now-deleted `_judge_code_level`.
13. Selection-parsing loop in `assign_codes` (`orchestrator.py:1458-1477`) mixes index/dedup validation, confidence-range validation, and threshold partitioning in one block — consider extracting `_partition_selections(selections, options, threshold)`.
14. `c.confidence` is assigned to four fields verbatim (`orchestrator.py:1492-1495`) with no local var/comment tying them together — a future "fix" to one could silently break the intentional mirroring documented on `AssignedCodeModel`.

### Confirmed clean (checked, no action needed)
- No dangling references to removed symbols (`JudgeResponse`, `build_judge_messages`, `build_pick_messages`, `parse_selected_indices`) anywhere in `src/` or `tests/`.
- `routes.py`'s empty-content short-circuit and `assign_codes`'s empty-framework guard are distinct conditions with no overlap/regression.
- `LLMResponseParseError` is caught only by the base `LLMError` handler in `app.py` (subclass match), so every other `response_model=` call site (analyze, hierarchical, summarize, sensitivity) is unaffected.
- `assign_codes` bypassing `_bounded_complete` matches existing convention (other single-call use cases do the same); not a new special case.

---
🤖 Generated by an automated 7-angle review (Claude Code)

olafdegraaff:
This is actually a good point:
"
No API contract break: response shape is unchanged, no symbols removed/renamed, and the new LLMResponseParseError inherits from LLMError so all other endpoints' error handling still works untouched.
The real issue: confidence_level_2, confidence_level_3, and confidence_aggregate now all just mirror the single self-reported confidence_level_1 value, instead of being independently judged / min()-aggregated like before. Any downstream consumer (EspoCRM dashboards, exports) that relied on level-divergence as a signal will silently get three identical numbers with no error thrown.
The old structural guarantee — a child code could only be selected if its ancestors first cleared the threshold — is gone, replaced by prompt wording only ("regardless of depth"). Combined with no ancestor/descendant dedup, the LLM can now legally return both a parent and a child path for the sa
"
I will give it a thought before deciding if this implementation makes the most sense.

PR #275, opened 2026-08-13
refactor(coding): reconcile one-shot assign_codes with the extracted LLMCallExecutor

### LINKED ISSUE #251: [SPIKE] SYSTEM_PROMPT assign codes
Currently the assign codes starts at level 1, then checks level 2 and then level 3. See if the performance improves if we let the LLM look at all possible combinations at once.

olafdegraaff:
closes #251 

## Summary
- Ports the one-shot code-classification `assign_codes` (from `feat/different-coding-method`) onto current `main`, which had since extracted the `LLMCallExecutor` collaborator (ADR-017).
- Drops the now-dead per-level pick/judge traversal methods that only existed to feed the old design, and rewires `assign_codes`'s two remaining scaffolding calls to `self._executor` instead of the removed private methods.
- Fixes two docs (`docs/architecture/03-components.md`, `docs/operations/settings-reference.md`) that still described a per-level judge inside `assign_codes` which had already stopped existing.

## Test plan
- [x] `make lint` passes (ruff, ty, import-linter contracts)
- [x] `make test` passes (624 tests)


src/qfa/services/coding_classifier.py
```diff
@@ -1,129 +1,115 @@
...
+class CodingResponse(BaseModel):
+    """Structured output for one-shot hierarchical code selection.
+
+    Confidence and explanation are deliberately absent: the pick step only
+    chooses *which* paths are in play. A separate per-level judge call (see
+    :func:`build_judge_messages`) — unchanged from the previous per-level
+    pick/judge design — scores and explains each one afterwards.
+    """
+
+    selected: list[int] = Field(
```
olafdegraaff, line 60: `CodingResponse.selected: list[int]` is strictly validated by pydantic, so one malformed element invalidates the *entire* list rather than just that element.

Confirmed: `CodingResponse.model_validate_json('{"selected": [0, "garbage", 2]}')` raises `ValidationError` for the whole payload. The deleted per-element-lenient `_parse_selected_indices` would have kept `[0, 2]`; now `except LLMResponseParseError: selected_indices = []` in `orchestrator.py` drops *all* selected indices, including the valid ones, producing `NO CODING APPLIED` for a record the model did partially classify correctly. No test covers a partially-valid list — only fully-valid/fully-invalid cases.

src/qfa/services/orchestrator.py
```diff
@@ -1339,44 +1348,91 @@ async def assign_codes(
...
+        try:
+            response = await self._llm.complete(
+                system_message=system_message,
+                user_message=anonymized_user_message,
+                tenant_id=request.tenant_id,
+                response_model=CodingResponse,
+                timeout=timeout,
+            )
+            selected_indices = response.structured.selected
+        except LLMResponseParseError:
```
olafdegraaff: This catch silently treats a pick-response schema-validation failure identically to a genuine "nothing relevant" judgment, with no logging to tell the two apart.

Elsewhere in this file, judge-call failures are logged with `error_class` before being swallowed (e.g. the `analyze` judge catch at line ~565, and the hierarchical leaf judge catch at line ~1018). This catch has no equivalent `logger.warning(...)`. If a provider starts emitting malformed `selected` payloads (model drift, provider change), every affected record silently gets `NO_CODING_NOTHING_RELEVANT_EXPLANATION` with zero telemetry to distinguish a real parsing regression from records that are genuinely uncodeable.

src/qfa/services/orchestrator.py
```diff
@@ -1339,44 +1348,91 @@ async def assign_codes(
...
-            accumulated_scores=[],
-            accumulated_explanations=[],
-            threshold=request.confidence_threshold,
-            tenant_id=request.tenant_id,
-            deadline=deadline,
-            candidates=candidates,
-            rejected=rejected,
-        )
+        seen_indices: set[int] = set()
+        for idx in selected_indices:
```
olafdegraaff: Judging multiple one-shot-selected paths (`for idx in selected_indices: await self._judge_selected_path(...)`) runs fully sequentially even though the paths are independent and could be judged concurrently (e.g. via `asyncio.gather`).

A record whose feedback spans several distinct topics can now select multiple paths in one pick call, and each is judged 1-3 levels deep in sequence, so latency scales linearly with the number of selected paths. Not a new issue introduced by this PR (the old per-level design had the same property), but the one-shot pick's wider selection surface makes multi-path selections more likely, so this loop becomes the larger share of latency going forward.

src/qfa/services/coding_classifier.py
```diff
@@ -1,129 +1,115 @@
...
+    any depth — every node in the framework (not just leaves) is its own
+    option, so a level-1-only or level-1+2 code can be selected directly
+    when nothing more specific fits.
+    """
 
+    path: tuple[tuple[str, str], ...]
 
-_SYSTEM = """You are a classification agent for feedback records from community members.
+    @property
+    def label(self) -> str:
```
olafdegraaff, line 30: `CodePathOption.label` here duplicates the exact same `" > ".join(name for _, name in path)` logic used to render rejected-candidate paths in `orchestrator.py`'s `_combine_rejected_explanations` (`f"{' > '.join(name for _, name in c.path)} — "`). Consider extracting one shared helper (e.g. `format_code_path(path) -> str`) so the LLM-facing options list and the EspoCRM-facing rejection explanation can't silently diverge in format if this rendering ever changes.

src/qfa/services/orchestrator.py
```diff
@@ -1339,44 +1348,91 @@ async def assign_codes(
...
-            path_ids_names=[],
-            accumulated_scores=[],
-            accumulated_explanations=[],
-            threshold=request.confidence_threshold,
-            tenant_id=request.tenant_id,
-            deadline=deadline,
-            candidates=candidates,
-            rejected=rejected,
-        )
+        seen_indices: set[int] = set()
```
olafdegraaff: Minor: `seen_indices: set[int]` + manual `if idx in seen_indices: continue` reimplements the order-preserving dedup idiom that this same PR deletes from `coding_classifier._parse_selected_indices` (which used `dict.fromkeys`). Not a bug, just an inconsistency — the codebase now does the same kind of dedup two different ways.

src/qfa/services/orchestrator.py
```diff
@@ -1427,6 +1484,82 @@ async def assign_codes(
...
+        Reproduces the previous per-level pick/judge design's judge step
+        exactly — same prompt, same score/explanation contract, same
+        early-stop-on-rejection behaviour — the only difference being that
+        the path being judged was already chosen in one shot rather than
+        picked one level at a time.
+        """
+        scores: list[float] = []
+        explanations: list[str] = []
+        hierarchy_path: list[tuple[str, str]] = []
+        for level_num, (code_id, name) in enumerate(path, start=1):
```
olafdegraaff: `code_id` is unpacked from `path` here but never used in the rest of `_judge_selected_path`. Harmless, but a linter configured to flag unused variables would catch this — consider `for level_num, (_, name) in enumerate(path, start=1):`.

olafdegraaff (commented):
Automated review (re-run after the PR's latest changes).

src/qfa/services/coding_classifier.py
```diff
@@ -1,129 +1,142 @@
...
+
+def flatten_coding_nodes(
+    nodes: list[CodingNode], _prefix: tuple[tuple[str, str], ...] = ()
+) -> list[CodePathOption]:
+    """Flatten a coding tree into one option per node, at every depth.
+
+    Pre-order: a parent immediately precedes its own children, so related
+    paths stay grouped together for the model.
+    """
+    options: list[CodePathOption] = []
```
mariushelf, line 43: ok, so we don't only have the leaves of the coding framework, but also all intermediate level 1 and level 2 nodes.

Do we need / want those?

(just gathering my thoughts)

src/qfa/services/coding_classifier.py
```diff
@@ -1,129 +1,142 @@
...
+    return options
+
+
+class CodingResponse(BaseModel):
+    """Structured output for one-shot hierarchical code selection.
+
+    Confidence and explanation are deliberately absent: the pick step only
+    chooses *which* paths are in play. A separate per-level judge call (see
+    :func:`build_judge_messages`) — unchanged from the previous per-level
+    pick/judge design — scores and explains each one afterwards.
```
mariushelf, line 57: Nit: "previous" = old implementation that (soon) no-one knows about anymore? Then this is a typical claude artifact -- claude likes to reference "previous" things that don't exist anymore.

src/qfa/services/orchestrator.py
```diff
@@ -1389,116 +1453,56 @@ async def assign_codes(
...
+        picked one level at a time.
+
+        Returns the scored path and whether it was rejected by
+        ``threshold``, so independently-selected paths can be judged
+        concurrently instead of one at a time.
+        """
+        scores: list[float] = []
+        explanations: list[str] = []
+        hierarchy_path: list[tuple[str, str]] = []
+        for level_num, (_, name) in enumerate(path, start=1):
```
mariushelf, line 1480: Nit: this works, but do we need 3 calls for this? Or could we do one call, and either only judge whether the final selection works, or instruct it to check all levels at once?

No action required now.

mariushelf (approved):
LGTM! Some thinking out loud/nit comments in the code, but approved without any required changes.

Issue #266, opened 2026-08-07
refactor(services): extract AnalyzeService from Orchestrator


mariushelf:
Part of the `Orchestrator` decomposition epic (#112). This is the largest extraction
(~700 lines) and lands last among the four, on a pattern already proven by the other
three.

Depends-on: #262

## Spec

**What:** Extract both analyze modes and their helpers into a single `AnalyzeService` in
`src/qfa/services/analyze.py`, add a `get_analyze_service` DI provider, and repoint the
analyze route (`src/qfa/api/routes.py`, ~lines 225–227, which selects between the two
modes) at it.

Members that move:

| Member | Current line | Mode |
|---|---|---|
| `analyze_bulk` | ~369 | bulk |
| `analyze_hierarchical` | ~491 | hierarchical |
| `_map_chunk` | ~906 | hierarchical |
| `_judge_chunk` | ~934 | hierarchical |
| `_reduce_partials` | ~988 | hierarchical |
| `_group_partials_to_budget` | ~1088 | hierarchical |
| `_coverage_weighted_mean` | ~1128 | hierarchical |
| `_is_retained_analyze_placeholder` | ~357 | both |
| `_ANALYZE_RETAINED_PLACEHOLDER_TYPES` | ~331 | both |

Plus the module-level analyze/judge helpers `AnalyzeJudgeResult` (~175),
`_parse_judge_quality_score` (~189), `_build_judge_system_message` (~201), and
`_SlotTiming` (~278), to the extent they are used only by the analyze path.

**Why:** `analyze_bulk` and `analyze_hierarchical` are two modes of one endpoint — the
route already selects between them on the request — and they share the
retained-placeholder rule (`_ANALYZE_RETAINED_PLACEHOLDER_TYPES`), which is a
defense-in-depth guardrail that must stay in one place. Splitting them into two services
would either duplicate that rule or require a third object to hold it, for no gain. They
therefore stay in one `AnalyzeService`, per the decision taken in #112.

`AnalyzeService` is the only use case that needs the `EmbeddingPort` — the hierarchical
path requires it, and `None` makes that path raise `AnalysisError` at request time. That
dependency becomes explicit on this one service's constructor instead of sitting on a
shared constructor that four other use cases do not need. This is precisely the payoff
the composition-only decision was made for.

## Acceptance criteria

- [ ] `src/qfa/services/analyze.py` defines `class AnalyzeService:` with **no base
[...]

Issue #264, opened 2026-08-07
refactor(services): extract SummarizeService from Orchestrator


mariushelf:
Part of the `Orchestrator` decomposition epic (#112).

Depends-on: #262

## Spec

**What:** Extract `Orchestrator.summarize_bulk` (~line 1140) and `Orchestrator.summarize`
(~line 1218) into `SummarizeService` in `src/qfa/services/summarize.py`, add a
`get_summarize_service` DI provider, and repoint both summarize route handlers
(`src/qfa/api/routes.py`, ~lines 316 and 383) at it.

**Why:** These two methods are the clearest natural pair in the class — both are
single-completion summarization over feedback records, differing only in bulk vs.
single-record shape, and neither is used by any other use case. Grouping them into one
service keeps their shared prompt and hyperlinking conventions co-located while removing
~150 lines from `Orchestrator`.

Note the module-level helper `_hyperlink_form_references` (~line 220) is used by the
summarize path; move it with the service unless another retained use case still needs it,
in which case leave it shared and note where it lands.

## Acceptance criteria

- [ ] `src/qfa/services/summarize.py` defines `class SummarizeService:` with **no base
      class**, taking an `LLMCallExecutor` as a constructor dependency.
- [ ] `summarize_bulk` and `summarize` are moved onto it and **deleted** from
      `Orchestrator`.
- [ ] Any module-level helper used only by these two methods (e.g.
      `_hyperlink_form_references`) moves with them; anything still shared with a
      retained use case stays put, and the PR description says which and why.
- [ ] `get_summarize_service` is added to `src/qfa/api/dependencies.py`.
- [ ] Both summarize route handlers type-annotate against `SummarizeService` and no
      longer take `Orchestrator`.
- [ ] `src/qfa/api/composition.py` constructs the service, sharing the one
      `LLMCallExecutor` instance.
- [ ] Tests for these two use cases move to `tests/services/test_summarize.py`,
      constructing the real `LLMCallExecutor` over `FakeLLMPort` / `FakeAnonymizer`. No
      test assertion is weakened or deleted in the move — in particular the
      output-language, hyperlink, and no-trailing-question cases are carried over intact.
- [ ] The existing e2e/API tests for both summarize endpoints pass **unmodified**.
- [ ] `make test` and `make lint` pass, including all three `import-linter` contracts.

mariushelf:
~Written by Claude, run via the agentic engineering loop

mariushelf:
[...]

PR #277, opened 2026-08-13
refactor(services): extract SensitivityService from Orchestrator

### LINKED ISSUE #263: refactor(services): extract SensitivityService from Orchestrator

mariushelf:
Committed, tree clean.

---

### What changed

**New service** — `src/qfa/services/sensitivity.py` defines `SensitivityService`, no base class, one constructor dependency (`LLMCallExecutor`). `detect_sensitive_content` plus its two prompt constants moved here and are **deleted** from `Orchestrator`, which drops from six public use cases to five.

**Executor grew three methods** (`src/qfa/services/llm_call_executor.py`) — the friction this "smallest extraction first" was meant to surface. The use case needs string-level PII handling and a single non-semaphore call, none of which #262 exposed:
- `anonymize_text(text)` — redact one assembled message (batch sibling of `anonymize_records`).
- `deanonymize_json(payload, mapping)` — the restore round trip, now owning the JSON escaping that was `Orchestrator._json_escape_mapping`. That private helper is deleted and the orchestrator's two remaining call sites (`summarize`, `summarize_bulk`) delegate here, so the escaping lives in one place.
- `complete(...)` — one deadline-bounded completion. `bounded_complete` is now literally "`complete` under a semaphore, with queue-wait timing" and delegates to it.

All three sit inside the remit ADR-017 decision 1 assigns the executor ("retry, deadline enforcement, token-limit check, anonymize/deanonymize"), and the next three extractions hit the identical shape at `summarize` / `summarize_bulk` / `assign_codes`.

**Wiring** — `get_sensitivity_service` added to `qfa/api/dependencies.py`; the `/v1/detect-sensitive` handler annotates against `SensitivityService` and no longer takes `Orchestrator`. `composition.build_services()` returns a frozen `ServiceGraph(orchestrator, sensitivity)` built over **one** shared `LLMCallExecutor`; `build_orchestrator` survives as a thin wrapper so scripts, notebooks and existing tests are untouched. The lifespan publishes `app.state.sensitivity_service`.

**Tests** — 4 tests moved verbatim in intent to `tests/services/test_sensitivity.py`, constructing the real executor over `FakeLLMPort`/`FakeAnonymizer`; no assertion weakened or dropped. Added: executor coverage for `complete` (deadline-derived timeout, expired deadline raises before the call) and the anonymise/deanonymise round trip including the quote-and-newline escaping case; composition coverage that the services share one executor instance; a lifespan test that every service reaches `app.state`.

**Docs** — `docs/architecture/03-components.md` (use-case table, new extracted-services table, expanded executor table, composition-root prose) and a note in `docs/development/implementing-a-new-endpoint.md` pointing new endpoints at the service+provider pattern instead of `get_orchestrator`.

### Verification
`make lint` clean (ruff, ty, all 3 import-linter contracts kept) · `make test` 621 passed · strict `make docs` (`-W`) builds.

### Impact / risk
No API change: response shape, status codes and error payloads for `/v1/detect-sensitive` are identical, and the endpoint's assertions are untouched. Risk is concentrated in two places rather than the new service: (1) `bounded_complete` now derives its timeout inside `complete` — same position relative to the semaphore acquire, but on the deadline-expired path `SlotTiming.call_seconds` is now populated instead of left at `0.0` (debug logging only, no test depended on it); (2) two `summarize` paths switched to `executor.deanonymize_json`, same escaping, same behaviour.

### One deviation from the acceptance criteria
The AC asked that existing API tests pass **unmodified**. One line of *fixture wiring* had to change: `tests/api/test_routes.py::test_detect_sensitive_502_analysis_error` reassigns `test_app.state.orchestrator` to inject its error fake, and the route no longer reads that slot — it now assigns `state.sensitivity_service`. Assertions are unchanged. `tests/api/conftest.py` points both slots at the same `FakeOrchestrator`, so every other sensitivity test (including the metadata-forwarding one that asserts on `state.orchestrator`) is byte-identical.

`spec/issue-263.md` checkboxes were left unticked — the spec is the signed-off record, not a working file.

Fixes #263

### DIFF
--- a/docs/architecture/03-components.md
+++ b/docs/architecture/03-components.md
@@ -97,7 +97,7 @@ the rationale, and flow/sequence diagrams.
 
 ## The orchestrator
 
-{py:class}`~qfa.services.orchestrator.Orchestrator` is one class with five async methods, each backing one HTTP endpoint:
+{py:class}`~qfa.services.orchestrator.Orchestrator` is one class with four async methods, each backing one HTTP endpoint:
 
 | Method | Endpoint | What it does |
 |---|---|---|
@@ -105,22 +105,40 @@ the rationale, and flow/sequence diagrams.
 | `analyze_hierarchical` | `POST /v1/analyze` (`mode=hierarchical`) | Embed -> cluster -> map -> reduce pipeline. Returns additional `confidence` and `coding_trends` fields. |
 | `summarize` | `POST /v1/summarize` | One LLM call. Per-record summaries with a self-evaluated quality score. |
 | `assign_codes` | `POST /v1/assign-codes` | Multiple LLM calls per record: pick + judge at each level of a hierarchical coding framework. |
-| `detect_sensitive_content` | `POST /v1/detect-sensitive` | One LLM call per record. Detects sensitive content and categorizes sensitivity types. |
+
+It is being decomposed into one service per use case by epic #112, so this
+table shrinks by one row per extraction and the class disappears entirely at
+the end of it. The extracted services are listed below.
 
 `/v1/summarize`, `/v1/assign-codes`, and `/v1/detect-sensitive` are non-bulk endpoints with per-record outputs. `/v1/analyze` is the bulk endpoint and returns one aggregate result per request (for both `mode=single_pass` and `mode=hierarchical`).
 
 Each method is pure use-case logic — no scope or correlation plumbing. `call_scope` is entered by a FastAPI dependency declared on the route (`Depends(call_scope_for(Operation.X))`), so by the time an orchestrator method runs `current_call_context` is already set. See [Cross-cutting concerns](04-crosscutting.md) for the full picture.
 
+### Extracted use-case services
+
+Each service below is a plain class in `qfa.services` that owns exactly one use
+case, receives the shared {py:class}`~qfa.services.llm_call_executor.LLMCallExecutor`
+as a constructor dependency, and is injected into its route by its own provider in
+`qfa.api.dependencies`. None of them has a base class — see
+[ADR-017](../adr/017-orchestrator-composition-only.md).
+
+| Service | Endpoint | Provider | What it does |
+|---|---|---|---|
+| {py:class}`~qfa.services.sensitivity.SensitivityService` | `POST /v1/detect-sensitive` | `get_sensitivity_service` | One LLM call per record. Detects sensitive content and categorizes sensitivity types. |
+
 ### The LLM-call executor
 
-The scaffolding every use case wraps its LLM calls in lives on one collaborator, {py:class}`~qfa.services.llm_call_executor.LLMCallExecutor`, which the orchestrator holds as `self._executor` and delegates to:
+The scaffolding every use case wraps its LLM calls in lives on one collaborator, {py:class}`~qfa.services.llm_call_executor.LLMCallExecutor`, which the orchestrator and every extracted service hold as `self._executor` and delegate to:
 
 | Method | Concern |
 |---|---|
 | `check_deadline_and_get_timeout(deadline)` | Derive the per-call timeout from the remaining request budget; raise {py:exc}`~qfa.domain.errors.AnalysisTimeoutError` when too little time is left |
 | `check_token_limit(system_message, user_message)` | Pre-flight token estimate; raise {py:exc}`~qfa.domain.errors.FeedbackTooLargeError` when over `LLM_MAX_TOTAL_TOKENS` |
 | `anonymize_records(records, anonymize)` | Redact each record's text, returning new records plus the merged restore mapping |
-| `bounded_complete(semaphore, …)` | One completion bounded by both a concurrency semaphore and the deadline; used by the hierarchical map/judge/reduce phases |
+| `anonymize_text(text)` | Redact one assembled message, returning the redacted text plus its restore mapping |
+| `deanonymize_json(payload, mapping)` | Restore redacted values inside a serialized JSON response, escaping them so the payload stays valid JSON |
+| `complete(…)` | One completion bounded by the deadline; used by the single-call use cases |
+| `bounded_complete(semaphore, …)` | `complete` run under a concurrency semaphore, with queue-wait timing; used by the hierarchical map/judge/reduce phases |
 
 It is a **plain concrete class** — not a Protocol, not a base class, and not declared in `qfa.domain.ports`. It wraps no external system, so it is not a port; and behaviour reuse in this codebase is always composition, so nothing inherits from it. Both points are decided in [ADR-017: Decompose the Orchestrator by composition only](../adr/017-orchestrator-composition-only.md), which is also why the composition root injects the executor rather than letting the service construct its own.
 
@@ -133,11 +151,11 @@ Tests construct the real executor over the existing `FakeLLMPort` / `FakeAnonymi
 `qfa.api.app.create_app()` builds the FastAPI instance; the `lifespan` context manager wires the dependency graph at startup. The wiring splits into two halves:
 
 - **Infrastructure half** (in `qfa.api.app`): load settings, build the base LLM client — plus a second one for judge calls when `JUDGE_LLM_MODEL` is set — create the async DB engine, wrap each LLM in {py:class}`~qfa.adapters.tracking_llm.TrackingLLMAdapter` for usage tracking, set up the auth adapter, build the embedder (logging its construction so operators see it on startup).
-- **Domain half** (in {py:func}`qfa.api.composition.build_orchestrator`): given settings plus the already-wrapped LLM and embedder, construct the {py:class}`~qfa.adapters.presidio_anonymizer.PresidioAnonymizer` and the {py:class}`~qfa.services.llm_call_executor.LLMCallExecutor`, register custom model prices with LiteLLM, and assemble the {py:class}`~qfa.services.orchestrator.Orchestrator`.
+- **Domain half** (in {py:func}`qfa.api.composition.build_services`): given settings plus the already-wrapped LLM and embedder, construct the {py:class}`~qfa.adapters.presidio_anonymizer.PresidioAnonymizer` and the {py:class}`~qfa.services.llm_call_executor.LLMCallExecutor`, register custom model prices with LiteLLM, and assemble every application service — returned together as a {py:class}`~qfa.api.composition.ServiceGraph` so they share the one executor.
 
-The lifespan then attaches the orchestrator, API keys, and the usage repository to `app.state` for the request lifecycle to read.
+The lifespan then attaches each service (`app.state.orchestrator`, `app.state.sensitivity_service`), the API keys, and the usage repository to `app.state` for the request lifecycle to read. One `app.state` slot per service is what lets each route's provider inject only the use case it needs.
 
-The split exists so callers outside the API server — scripts, notebooks, ad-hoc evaluation harnesses — can construct an `Orchestrator` over a plain LLM client with a single call:
+The split exists so callers outside the API server — scripts, notebooks, ad-hoc evaluation harnesses — can construct an `Orchestrator` over a plain LLM client with a single call ({py:func}`~qfa.api.composition.build_orchestrator` is the narrow wrapper over `build_services` for exactly that case):
 
 ```python
 from qfa.api.composition import build_orchestrator
--- a/src/qfa/services/llm_call_executor.py
+++ b/src/qfa/services/llm_call_executor.py
@@ -1,10 +1,11 @@
 """Shared LLM-call scaffolding for the application services.
 
 Every use case in :mod:`qfa.services` wraps its LLM calls in the same
-four concerns: redact PII before the call, derive a per-call timeout from
-the request deadline, guard the token budget, and bound concurrent calls
-with a semaphore. :class:`LLMCallExecutor` owns those four and nothing
-else, so a use-case service can be read without them.
+four concerns: redact PII before the call (and restore it afterwards),
+derive a per-call timeout from the request deadline, guard the token
+budget, and bound concurrent calls with a semaphore.
+:class:`LLMCallExecutor` owns those four and nothing else, so a use-case
+service can be read without them.
 
 Per ADR-017 this is a plain concrete class, deliberately **not** a
 Protocol and **not** a base class:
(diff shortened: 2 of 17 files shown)


mariushelf:
Build complete. 

mariushelf:
Builder committed 1 of the 10 tasks its own plan (`tasks/plan.md`) declares. This is a signal for the review round, not a verdict: a short count *can* mean the build was cut off and the remaining tasks were never built, but a verification-only task commits nothing and one commit can cover several tasks. Check the diff against the spec's acceptance criteria before trusting the summary above.

tests/api/test_routes.py
```diff
@@ -866,7 +866,7 @@ async def test_summary_returns_configured_result(self, test_app):
 
     @pytest.mark.asyncio
     async def test_detect_sensitive_502_analysis_error(self, test_app):
-        test_app.state.orchestrator = FakeOrchestrator(
+        test_app.state.sensitivity_service = FakeOrchestrator(
```
mariushelf, line 869: **minor** — this slot rewire is right, but the sibling test at `tests/api/test_routes.py:348` (`test_detect_sensitive_forwards_metadata`) was left reading the *other* slot:

```python
assert test_app.state.orchestrator.last_detect_sensitive_request is not None
```

The route now resolves `state.sensitivity_service`, so that assertion only holds because `tests/api/conftest.py:330-332` aliases both slots to the same `fake_orchestrator` object. It passes today (I ran it), but it asserts on a collaborator the handler no longer touches, and it will start failing for a reason unrelated to what it tests as soon as `FakeOrchestrator` is split per service in #264–#267. Reading `test_app.state.sensitivity_service.last_detect_sensitive_request` here would keep the test pointed at the object under test.

tests/services/test_sensitivity.py
```diff
@@ -0,0 +1,177 @@
...
+    SensitivityAnalysisResultModelList,
+)
+from qfa.domain.sensitivity_types import SensitivityType
+from qfa.services.llm_call_executor import LLMCallExecutor
+from qfa.services.sensitivity import SensitivityService
+from qfa.settings import OrchestratorSettings
+
+# Reuse the doubles the orchestrator suite already ships rather than growing a
+# second, drifting pair (ADR-017).
+from .test_orchestrator import FakeAnonymizer, FakeLLMPort
```
mariushelf, line 29: **minor** — reusing the existing doubles rather than growing a second pair is the right call (ADR-017 decision 3), but importing them *from a sibling test module* couples the new suite to the one it was extracted from, and `test_orchestrator.py` is scheduled for deletion in #267 — at which point `FakeAnonymizer` / `FakeLLMPort` disappear from under `test_sensitivity.py`.

This PR is explicitly the pattern the three larger extractions copy, so this import is the shape #264–#266 will replicate three more times. Promoting the two doubles to `tests/services/conftest.py` (or a `tests/services/doubles.py`) now makes each extracted suite stand on its own and removes one item from #267's cleanup.

src/qfa/services/sensitivity.py
```diff
@@ -0,0 +1,113 @@
...
+        -------
+        SensitivityAnalysisResultModel
+            The sensitivity analysis result for the feedback record.
+        """
+        system_message = _DEFAULT_SENSITIVITY_DETECTION_PROMPT
+        user_message = build_feedback_record_envelope(
+            request.feedback_record, include_metadata=True, include_id=True
+        )
+
+        anonymized_user_message, anonymization_mapping = self._executor.anonymize_text(
```
mariushelf, line 88: **nit** (no action needed, recorded so the next extraction does not have to re-derive it) — the move reorders two steps: the original `Orchestrator.detect_sensitive_content` called `check_deadline_and_get_timeout` *first*, before building the envelope and anonymising; here the deadline check happens last, inside `executor.complete`.

Same exception (`AnalysisTimeoutError`), same status code, same payload — the only difference is that an already-expired request now pays a Presidio pass before raising. Not observable through the API, and folding the check into `complete()` is what makes the executor reusable, so I read this as an accepted consequence rather than a regression.

src/qfa/services/llm_call_executor.py
```diff
@@ -109,6 +125,18 @@ def anonymize_records(
             new_records.append(record.model_copy(update={"content": redacted}))
         return tuple(new_records), merged
 
+    def anonymize_text(self, text: str) -> tuple[str, dict[str, str]]:
```
mariushelf, line 128: **nit** — `anonymize_text(text)` takes no `anonymize: bool`, while its sibling `anonymize_records(records, anonymize)` does. That is faithful to the code being moved (detect-sensitive always anonymised, unconditionally), so nothing is broken here.

Worth a docstring line all the same: the asymmetry now sits in the shared collaborator that #264–#266 will reach for, and the next author assembling a single message has no way to tell whether the missing flag is a deliberate "this path always redacts" or an oversight.

docs/architecture/03-components.md
```diff
@@ -97,30 +97,48 @@ the rationale, and flow/sequence diagrams.
 
 ## The orchestrator
 
-{py:class}`~qfa.services.orchestrator.Orchestrator` is one class with five async methods, each backing one HTTP endpoint:
+{py:class}`~qfa.services.orchestrator.Orchestrator` is one class with four async methods, each backing one HTTP endpoint:
```
mariushelf, line 100: **nit** — the decrement is faithful (one row removed, `five` → `four`), but it carries forward an off-by-one that was already there: `Orchestrator` still has *five* public async methods — `analyze_bulk`, `analyze_hierarchical`, `summarize_bulk`, `summarize`, `assign_codes` — and the table has always omitted `summarize_bulk` while naming `analyze_bulk` as `analyze`.

Pre-existing, not introduced here, but this PR edits exactly this sentence, which makes it the cheapest moment to fix: `five async methods` with a `summarize_bulk` / `POST /v1/summarize-bulk` row added.

mariushelf (commented):
## Review: intent vs. behaviour

The extraction itself is faithful and complete. `SensitivityService` is a plain class with
no base class and exactly one constructor dependency, `detect_sensitive_content` is gone
from `Orchestrator` (along with its prompt constants and the now-unused `json` import),
`get_sensitivity_service` is in `dependencies.py`, the route annotates against the service
and no longer names `Orchestrator`, and `build_services` hands both the orchestrator and
the new service the *same* executor instance — asserted by identity in
`test_composition.py` and again at the lifespan level in `test_lifespan.py`. All four
`TestDetectSensitiveContent` tests moved to `tests/services/test_sensitivity.py`
assertion-for-assertion over the real `LLMCallExecutor` built on `FakeLLMPort` /
`FakeAnonymizer`. Every acceptance criterion is met.

Verified locally on `3619a2b`:

- `make test` — 621 passed, 42 deselected.
- `make lint` — ruff, `ty`, and `lint-imports` clean; **3 contracts kept, 0 broken**.
- `make docs` — builds clean, `qfa.services.sensitivity` picked up by the apidoc pass.

### On the "1 of 10 tasks" note

The short commit count is not a truncated build. One commit covers the whole extraction,
and the diff is if anything *larger* than the spec's framing (+714/−278 against a spec
that describes a ~50-line move):

- `LLMCallExecutor` gains `complete`, `anonymize_text` and `deanonymize_json`, and
  `bounded_complete` is rewritten to delegate to `complete`.
- `_json_escape_mapping` moves off `orchestrator.py` onto the executor, and the two
  remaining call sites in `summarize` / `summarize_bulk` are rewired to
  `deanonymize_json`.
- `composition.py` grows `build_services` + a `ServiceGraph` dataclass, with
  `build_orchestrator` retained as a thin wrapper for scripts/notebooks.

ADR-017 decision 1 names "anonymize/deanonymize" and "deadline enforcement" as exactly the
scaffolding the executor is meant to own, so all three executor additions sit inside its
stated remit. I checked the two behaviour-preservation questions that scope raises:

1. `deanonymize_json` runs on the *executor's* anonymiser, not the orchestrator's. Every
   construction path passes one instance to both (`composition.py:280-293`, and
   `Orchestrator.__init__:409` default-builds the executor over its own `anonymizer`), so
   the rewiring is a no-op. It would stop being one if a caller ever passed an executor
   built over a different anonymiser — no such call site exists today.
2. Moving the deadline check inside `bounded_complete`'s `try` means `timing.call_seconds`
   is now populated on the `AnalysisTimeoutError` path. Its only consumers are the
   post-success debug logs at `orchestrator.py:696` and `:766`, which never run on that
   path, so nothing observable changes.

### Findings

Five, all non-blocking — two minor, three nits, filed inline. Nothing that must change
before a human reads this PR.

### For the human reviewer

Two calls I deliberately did not make on your behalf:

- **`build_services` + `ServiceGraph` as the epic's shape.** The spec asked only that
  `composition.py` construct the service over the shared executor. This PR answers with an
  aggregate return type and one `app.state` slot per service. It is sound and it is the
  first of four, so #264–#266 will copy it — worth ratifying now rather than after three
  more services have been threaded through it.
- **Whether the executor/`summarize` changes belong in this PR.** They are
  behaviour-preserving (checked above) and within ADR-017's remit, but they are scope this
  ticket did not ask for.

One narrow caveat on AC 7 ("existing e2e/API tests pass **unmodified**"): every assertion,
response shape, status code and error payload is untouched, and every sensitivity test
except one passes with no edit at all. The exception is a single fixture-wiring line in
`test_detect_sensitive_502_analysis_error`, which installs its error-raising fake on
`state.sensitivity_service` instead of `state.orchestrator`. That is an unavoidable
consequence of AC 4 — the route no longer reads the orchestrator slot — so I read the
criterion as met rather than sending it back; flagging it so the deviation is on the
record rather than inferred. Separately, the `-m e2e` tier is excluded from `make test` by
`addopts` and cannot run here (no Docker/Postgres), but no file under `tests/e2e/`
references the sensitivity endpoint, so `tests/api/test_routes.py` is its full coverage.

**Verdict: approve.**

mariushelf:
Review iteration 1: **approve** — Faithful, complete extraction of `detect_sensitive_content` into `SensitivityService`. All 8 acceptance criteria met, verified locally on 3619a2b: `make test` 621 passed / 42 deselected, `make lint` clean with 3/3 import-linter contracts kept, `make docs` builds clean and picks up the new module. The service is a plain class with no base class and exactly one constructor dependency; the method (plus its prompt constants and the now-unused `json` import) is gone from `Orchestrator`; `get_sensitivity_service` is in `dependencies.py`; the route annotates against the service and no longer names `Orchestrator`; `build_services` hands both objects the same executor, asserted by identity at both the composition and lifespan levels; all four moved tests run assertion-for-assertion over the real `LLMCallExecutor` on `FakeLLMPort`/`FakeAnonymizer`.

The loop's "1 of 10 tasks" note is a false alarm: one commit covers the whole extraction and the diff is larger than the spec's ~50-line framing (+714/-278), not truncated. Beyond the spec it adds `complete`/`anonymize_text`/`deanonymize_json` to `LLMCallExecutor`, rewrites `bounded_complete` to delegate, moves `_json_escape_mapping` onto the executor and rewires `summarize`/`summarize_bulk` to `deanonymize_json`, and introduces `build_services` + a `ServiceGraph` dataclass with `build_orchestrator` kept as a thin wrapper. ADR-017 decision 1 names anonymize/deanonymize and deadline enforcement as the executor's remit, so those additions are in scope. I checked the two behaviour-preservation risks that scope raises and both are no-ops: `deanonymize_json` runs on the executor's anonymiser, which is the same instance in every construction path (composition.py:280-293 and Orchestrator.__init__:409); and moving the deadline check inside `bounded_complete`'s try now populates `timing.call_seconds` on the timeout path, whose only consumers are post-success debug logs that never run there.

Five findings, none gating: two minor (a sibling route test left asserting through `state.orchestrator` when the route reads `state.sensitivity_service`; test doubles imported from a test module due for deletion in #267) and three nits.


mariushelf:
Could not verify: Whether `build_services` + `ServiceGraph` (one aggregate return, one `app.state` slot per service) is the shape the epic wants. The spec asked only that `composition.py` construct the service over the shared executor; this PR answers with a design that #264-#266 will copy three more times. Sound as written, but ratifying it is the epic owner's call and is cheaper now than after three more services are threaded through it.; Whether the executor additions (`complete`, `anonymize_text`, `deanonymize_json`) and the `summarize` / `summarize_bulk` rewiring belong in this PR or in their own. I verified they are behaviour-preserving and within ADR-017 decision 1's stated remit, but the scope judgement is the human's.; The `-m e2e` tier could not be exercised: it is excluded from `make test` by `addopts`, and this environment has no Docker or Postgres (all 21 e2e tests skip on the DB probe). No file under `tests/e2e/` references the sensitivity endpoint, so `tests/api/test_routes.py` appears to be its full coverage, but a reviewer with a database should confirm.


olafdegraaff:
LGTM

olafdegraaff (approved):

Issue #112, opened 2026-05-15
Refactor `orchestrator.py`


teeuwski:
### **User Story**

As a developer
I want to be able to quickly grasp the `orchestrator` and its different components.
So that I can add/change functionality easily.

## Spec

**What:** Decompose `qfa.services.orchestrator.Orchestrator` (1626 lines, 6 unrelated
public methods) into per-use-case application services that share behaviour by
**composition only**, and remove the `Orchestrator` class entirely — each route handler
depends on the single service it actually needs.

**Why:** `Orchestrator` is a god-class. It has grown from 723 lines (at the 2026-05-08
architecture review) to 1626 lines, absorbing an entire hierarchical map/judge/reduce
analysis mode as more methods on the same class. Its six public methods
(`analyze_bulk`, `analyze_hierarchical`, `summarize_bulk`, `summarize`, `assign_codes`,
`detect_sensitive_content`) share a constructor but little else, so every reader must
traverse the whole file to change any one of them, and no use case can be tested or
evolved in isolation.

This is the escape valve [ADR-011](docs/adr/011-drop-orchestrator-port.md) explicitly
anticipated — *"extracting one into its own class when it grows enough to earn its own
collaborators"* — and the change
[the 2026-05-08 review recommended](docs/architecture-review-2026-05-08.md) (§ "What a
green-field redo would change", item 1). Its
[addendum](docs/architecture-review-2026-05-08-addendum.md) is explicit that this is
*"anticipated by ADR-011, not contrary to it"* and that the extraction is "just past
time."

### Architecture decisions (settled in triage, 2026-08-07)

Three decisions were taken during triage and are **not** open for re-litigation by the
implementing agents. The ADR child ticket records them.

1. **Composition only — no shared base class among the new services.** The shared
   LLM-call scaffolding (retry, deadline enforcement, token-limit check,
   anonymize/deanonymize) is extracted into a single collaborator object,
   `LLMCallExecutor`, which each service receives as a constructor dependency and
   delegates to. A shared abstract base class was considered and rejected: it would make
   each service's real dependencies implicit in a common `__init__` (some need an
   embedder, most do not) and reintroduce the inheritance-for-reuse style this codebase
   has consistently avoided.

2. **`Orchestrator` is deleted, not kept as a facade.** Each route handler gets its own
   DI provider (`get_analyze_service`, `get_coding_service`, …) and type-annotates
[...]

PR #278, opened 2026-08-13
refactor(services): extract SummarizeService from Orchestrator

### LINKED ISSUE #264: refactor(services): extract SummarizeService from Orchestrator

mariushelf:
### What changed

**New `src/qfa/services/summarize.py`** — `class SummarizeService:` with **no base class** (ADR-017), holding `summarize_bulk` and `summarize`, moved from `Orchestrator` and deleted there. Its constructor names only what these two use cases need: `llm`, `anonymizer`, `executor: LLMCallExecutor` (required), `judge_llm=None` — no embedder, no `max_total_tokens`, since neither path runs the token-budget guard. With this, `Orchestrator` holds no use case at all; #267 deletes the class.

**Helper split** (AC bullet 3 — which moved and why):

- *Moved with the service* (used only by these two methods): `_DEFAULT_SUMMARIZATION_PROMPT`, `_DEFAULT_AGGREGATE_SUMMARIZATION_PROMPT`, `_JUDGE_PROMPT`, `_build_judge_system_message`, `_parse_judge_quality_score`.
- *Stayed shared*: `hyperlink_form_references` (`qfa.services.record_links`, also used by `analyze_bulk` and the hierarchical reduce) and `JUDGE_USER_MESSAGE` (`qfa.services.prompts`, also used by the analyse and hierarchical leaf judges). Both already have a home outside `Orchestrator` on `main`, so this PR imports them from there — the service does not reach into the shrinking god class for anything.

**Composition/wiring:** `build_services` gains a `summarize` field on `ServiceGraph`, constructed over the *same* `LLMCallExecutor` and `PresidioAnonymizer` as every other service. The lifespan publishes `app.state.summarize_service`; `get_summarize_service` is added to `qfa.api.dependencies`; both summarize handlers now take `SummarizeService = Depends(get_summarize_service)` and no longer see `Orchestrator`. Naming follows the convention #263–#266 established — `ServiceGraph.summarize` / `app.state.summarize_service` / `get_summarize_service` — rather than introducing a fourth spelling.

**Tests:** the two use cases' tests moved to `tests/services/test_summarize.py` (output-language, hyperlink, no-trailing-question, judge-parsing and JSON-escape-deanonymisation cases carried over intact), driven over the **real** `LLMCallExecutor` on `FakeLLMPort`/`FakeAnonymizer`. Because those were the last tests left in `tests/services/test_orchestrator.py`, that module became `test_summarize.py` wholesale — it also ships the shared `FakeLLMPort`/`FakeAnonymizer` doubles, so the four sibling suites that import them (`test_analyze`, `test_coding`, `test_sensitivity`, `test_llm_call_executor`) are repointed one line each.

New assertions: `summarize` in the shared-executor / shared-anonymiser identity tests and a both-connections test in `test_composition.py`, and `app.state.summarize_service` publication in `test_lifespan.py`. The cross-service judge-routing tests stayed in their own module (their subject is the cross-service split, and splitting them per service is how a call site quietly stops being checked) but now drive the service.

No test file for either summarize endpoint was modified — `tests/api/conftest.py` resolves the new provider through a `dependency_overrides` entry reading `state.orchestrator` per request, so the dozen route tests that swap in an error-raising fake after fixture setup keep working untouched. The trade-off is that no route test resolves the real provider; the comment there now names `test_every_service_is_published_on_app_state` as the compensating check.

**Dead code dropped:** `summarize_bulk` accumulated a `total_cost` that nothing read (cost is summed per client by `TrackingLLMAdapter`). It was already unread on `main`; a move is the cheap moment to drop it. The judge-routing test whose docstring described that accumulation now describes what it actually asserts — one call on each client.

**Docs:** `docs/architecture/03-components.md` (owning service per method; the second service table now maps service→provider instead of repeating every description from the table above), `01-architecture-style.md`, `docs/development/implementing-a-new-endpoint.md` and the `AGENTS.md` architecture note.

### Verification

`make test` — 651 passed, 42 deselected, exit 0. `make lint` — ruff, `ty`, and all three import-linter contracts pass (3 kept, 0 broken). `make docs` builds clean.

### Impact / risk

Behaviour-preserving refactor: prompts, call order, judge routing, anonymisation round-trip and both endpoints' responses are unchanged. Risk concentrates in the wiring, not the logic — a service the lifespan forgets to publish would 500 on first request, which is why `test_every_service_is_published_on_app_state` asserts it at startup over the real composition path. Anything constructing an `Orchestrator` directly and calling `.summarize*` breaks loudly with `AttributeError`; the repo has no such caller left. The 42-test integration/e2e tier could not run here (no Docker/Postgres in this environment); its summarize tests are unmodified.

Fixes #264


mariushelf:
Build complete. 

mariushelf:

mariushelf (commented):
## Review — intent vs. behaviour (fresh context)

**Verdict: approve.** The extraction does what #264 asked, the moved logic is byte-identical to what it replaced, and every acceptance criterion checks out. Findings below are all non-blocking.

### On the loop's "1 of 9 tasks committed" note

Checked, and it's a false alarm of the benign kind: the single commit `18c497c` carries the whole scope — new service, orchestrator deletions, DI provider, composition graph, both route handlers, the test move, and the doc updates (16 files, +1245/−768). Nothing in the plan is missing from the diff.

### What I verified

| Criterion | Status | Evidence |
|---|---|---|
| `SummarizeService` in `services/summarize.py`, **no base class**, `LLMCallExecutor` ctor dep | met | `summarize.py:135` `class SummarizeService:`; `executor: LLMCallExecutor` is a required positional param (`:168`) |
| `summarize_bulk` / `summarize` moved **and deleted** from `Orchestrator` | met | −150 lines of use-case body from `orchestrator.py`; the 6 now-unused model imports and `build_feedback_records_envelope` dropped with them |
| Shared helpers stay put; PR says which and why | met | `_hyperlink_form_references` (still used at `orchestrator.py:454`, `:826`), `_json_escape_mapping` (`:1237`), `_JUDGE_USER_MESSAGE` (`:470`, `:923`) retained and imported; the three summarize-only ones (`_DEFAULT_*_PROMPT`, `_JUDGE_PROMPT`, `_build_judge_system_message`, `_parse_judge_quality_score`) moved. PR body states the split |
| `get_summarize_service` in `api/dependencies.py` | met | `dependencies.py:36` |
| Both handlers annotate `SummarizeService`, no `Orchestrator` | met | `routes.py:266`, `:341` — parameter, annotation, docstring and call site all repointed |
| `composition.py` constructs it over the **one** `LLMCallExecutor` | met | single `executor = LLMCallExecutor(...)` handed to both; asserted by identity (`is`) in `test_services_share_one_executor` and `test_services_share_one_anonymizer`, not by type |
| Tests moved to `tests/services/test_summarize.py` over the real executor; no assertion weakened | met | see below |
| Existing e2e/API tests for both endpoints pass unmodified | met | `tests/api/test_routes.py` and `tests/e2e/test_orchestrator_e2e.py` untouched; API tier ran green. Caveat under *Could not verify* |
| `make test` / `make lint` incl. all three import-linter contracts | met | reproduced locally: **620 passed, 42 deselected**; ruff + `ty` clean; **3 contracts kept, 0 broken** |

**On the test move specifically** — I didn't take "carried over intact" on faith. I extracted all 14 test bodies from `origin/main:tests/services/test_orchestrator.py` and diffed them against the new module mechanically. Every one differs only by swapping the 6-line `Orchestrator(...)` construction for `_build_service(...)` and `orch.` for `service.`; not a single assertion, `pytest.raises` match, or fixture value changed. The three the ticket calls out by name are all there and intact: `test_output_language_instructs_aggregate_system_message` / `test_no_language_directive_when_output_language_unset`, `test_hyperlinks_form_reference_in_summary` / `test_no_hyperlink_without_base_url`, and both `*_forbids_trailing_questions`. `_build_service` builds the **real** `LLMCallExecutor` over `FakeLLMPort`/`FakeAnonymizer` — no fake executor — and threads `max_total_tokens` through so `TestTokenLimit` still asserts a large payload *is* forwarded.

One thing worth calling out as a genuine improvement rather than a regression: `main` had two same-named `test_no_hyperlink_without_base_url` tests, and the one sitting inside `TestSummarizeBulkHyperlinks` at `test_orchestrator.py:854` actually drove `analyze_bulk`. That copy stayed with the analyze suite (`test_orchestrator.py:900`) and the summarize module got the version that exercises `summarize_bulk`, so the analyze-side coverage is retained and the summarize-side case now tests what its name says.

### Notes for the human reviewer (not filed inline — outside the diff)

- `docs/operations/settings-reference.md:23-25` still describes the four judge call sites as all living on the orchestrator. This PR updated the parallel sentence in `docs/architecture/03-components.md` but not this one. It isn't *wrong* — the sentence names methods, and `JUDGE_LLM_*` still routes exactly those four — so it's a nit, not doc rot. (Same file uses `summarize_aggregate` for the method now called `summarize_bulk`; that inconsistency predates this PR and is repo-wide.)
- `build_orchestrator` now constructs a `SummarizeService` it discards, for scripts and notebooks that only want the orchestrator. Free in practice (no I/O in that constructor) and it buys a single wiring path, which the PR explicitly tests for (`test_build_orchestrator_returns_the_graphs_orchestrator`). Mentioning it only so it isn't mistaken for an oversight.

### Could not verify

- The 42-test integration/e2e tier needs Docker/Postgres, absent here, so `tests/e2e/test_orchestrator_e2e.py::TestRequestIdEqualsCallId::test_fanned_out_llm_calls_share_one_call_id_equal_to_request_id` — the one e2e test that hits `/v1/summarize-bulk` and asserts ≥2 tracked LLM-call rows — did not execute. I traced why it should pass unchanged (`tests/e2e/conftest.py:174` boots the real `create_app`, so the real lifespan publishes `app.state.summarize_service`, and `test_lifespan.py` asserts the service holds the same `TrackingLLMAdapter`-wrapped clients as the orchestrator, which is what keeps judge calls billed). Worth a glance at the CI run before merge rather than trusting that trace.
- Whether the cross-service judge-routing tests belong in `test_orchestrator_judge_routing.py` or should follow the use case into `test_summarize.py` is a taste call. The PR argues for keeping them together — "splitting them per service is exactly how one call site quietly stops being checked" — which I find persuasive, but the module name now undersells its scope.

src/qfa/services/summarize.py
```diff
@@ -0,0 +1,328 @@
...
+
+        judge_timeout = self._executor.check_deadline_and_get_timeout(deadline)
+        judge_response = await self._judge_llm.complete(
+            system_message=judge_system,
+            user_message=_JUDGE_USER_MESSAGE,
+            tenant_id=request.tenant_id,
+            response_model=str,
+            timeout=judge_timeout,
+        )
+        total_cost += judge_response.cost
```
mariushelf: **nit (pre-existing, carried over):** `total_cost` is assigned on L222 and accumulated here, but never read again — `AggregateSummaryResultModel` gets `quality_score` set, not a cost, and the method returns a `model_copy` that never touches it. It was already dead in `Orchestrator.summarize_bulk`, so keeping it verbatim is the right call for a behaviour-preserving extraction; flagging it because moving the code into a fresh file is the cheap moment to drop it, and because `tests/services/test_orchestrator_judge_routing.py` still carries the comment "the total the service accumulated covers both" over an assertion that only counts calls — the accumulation it describes is unobservable.

src/qfa/services/summarize.py
```diff
@@ -0,0 +1,328 @@
...
+    SummaryRequestModel,
+    SummaryResultModel,
+)
+from qfa.domain.ports import AnonymizationPort, LLMPort
+from qfa.services.llm_call_executor import LLMCallExecutor
+from qfa.services.orchestrator import (
+    _JUDGE_USER_MESSAGE,
+    _hyperlink_form_references,
+    _json_escape_mapping,
+)
```
mariushelf: **nit:** this is exactly the split the spec asked for (`_hyperlink_form_references` / `_json_escape_mapping` / `_JUDGE_USER_MESSAGE` each still have a retained user in `Orchestrator`, verified: `orchestrator.py:454`, `:826`, `:1237`, `:470`, `:923`), and the module docstring is explicit that it is temporary. Worth noting the direction it leaves behind: the *new* service now depends on three underscore-private names in the *old* god-class, so #267 cannot delete `Orchestrator` without first rehoming them. A one-line note in #267 (or a `qfa.services.prompts` / small shared module landing spot named up front) would keep that from being rediscovered later.

tests/api/conftest.py
```diff
@@ -333,6 +334,16 @@ def test_app(fake_orchestrator, fake_auth_orchestrator):
...
 
+    # The summarize routes resolve their own service since #264 extracted
+    # SummarizeService, but ``FakeOrchestrator`` still implements every use
+    # case, so one fake keeps standing in for all of them. Resolved through
+    # an override rather than ``app.state.summarize_service`` so it is read
+    # per request: a dozen route tests swap ``state.orchestrator`` for an
+    # error-raising fake *after* this fixture ran, and they must keep
+    # driving the summarize endpoints without knowing which service class
+    # is behind them.
+    app.dependency_overrides[get_summarize_service] = lambda: app.state.orchestrator
```
mariushelf, line 357: **minor:** the override keeps the ~dozen route tests working untouched (that part is well judged, and the comment explains it), but it also means **no** API-level test resolves the real `get_summarize_service` or drives a `SummarizeService` through a route: `FakeOrchestrator` is not a `SummarizeService`, and the provider under test is replaced wholesale. So a signature drift between the route call site and `SummarizeService.summarize_bulk` would pass the route tier and only surface in the e2e tier (which needs Postgres and doesn't run in the default `make test` selection).

The wiring itself *is* covered — `test_every_service_is_published_on_app_state` and the three `TestBuildServiceGraph` identity tests are the right assertions, and I'd rather have them than a rewritten route tier. A cheap top-up would be one route test that leaves the override in place but points it at a `MagicMock(spec=SummarizeService)`, so the fake is conformance-checked against the real class at least once. Non-blocking.

src/qfa/api/composition.py
```diff
@@ -164,21 +174,39 @@ def register_custom_model_prices() -> None:
...
+
+    One field per application service. Epic #112 adds a field here per
+    extraction and removes ``orchestrator`` in #267; until then the two
+    coexist, which ADR-017 anticipates. All of them share the one
+    :class:`~qfa.services.llm_call_executor.LLMCallExecutor`
+    :func:`build_service_graph` builds.
+    """
+
+    orchestrator: Orchestrator
+    summarize: SummarizeService
```
mariushelf: **nit:** the field is `summarize`, the `app.state` attribute the lifespan publishes is `summarize_service`, and the provider is `get_summarize_service`. Three names for one object across `composition.py` → `app.py` → `dependencies.py`. Since epic #112 adds a field here per extraction, picking one convention now (`summarize_service` reads unambiguously next to `orchestrator`) keeps the next four from each choosing their own.

mariushelf:
Review iteration 1: **approve** — Behaviour-preserving extraction that does what #264 asked. All nine acceptance criteria are met, verified against the diff rather than the PR description. `SummarizeService` is a plain class with no base class taking a required `LLMCallExecutor`; both methods are deleted from `Orchestrator` along with their now-unused imports; the three helpers still used by retained use cases stay in `orchestrator.py` (confirmed live callers at :454, :826, :1237, :470, :923) and the PR body states the split; `get_summarize_service` exists; both handlers annotate `SummarizeService`; `build_service_graph` builds one executor and one anonymiser and hands the same instances to both services, asserted by identity rather than type. I mechanically extracted all 14 moved test bodies from `origin/main:tests/services/test_orchestrator.py` and diffed them against the new module: every one differs only by swapping the `Orchestrator(...)` construction for `_build_service(...)` and the receiver name — no assertion, `pytest.raises` match, or fixture value changed, and the ticket's three named cases are intact. `_build_service` uses the real `LLMCallExecutor` over `FakeLLMPort`/`FakeAnonymizer`. Reproduced `make test` (620 passed, 42 deselected) and `make lint` (ruff + `ty` clean, 3 import-linter contracts kept, 0 broken). The loop's "1 of 9 tasks" note is a false alarm — commit 18c497c carries the entire scope. Docs are updated thoroughly and accurately, including the anchor rename. One incidental improvement: `main` had two same-named `test_no_hyperlink_without_base_url` tests and the copy inside `TestSummarizeBulkHyperlinks` actually drove `analyze_bulk`; analyze-side coverage stayed put and the summarize module got a version that tests what its name says. Five findings, none gating: one minor (the `tests/api/conftest.py` override means no route-level test exercises `SummarizeService`) and four nits.


mariushelf:
Could not verify: The 42-test integration/e2e tier requires Docker/Postgres, absent in this environment, so tests/e2e/test_orchestrator_e2e.py::TestRequestIdEqualsCallId::test_fanned_out_llm_calls_share_one_call_id_equal_to_request_id (the only e2e test hitting /v1/summarize-bulk) did not execute. Its file is unmodified and it boots the real create_app, whose lifespan publication of app.state.summarize_service is directly asserted by test_lifespan.py — but the actual run needs a CI check before merge.; Whether the cross-service judge-routing tests should stay in tests/services/test_orchestrator_judge_routing.py or follow the use case into test_summarize.py is a taste call for the human reviewer. The PR's argument for keeping them together is reasonable, but the module name now undersells its scope.


mariushelf:
`main` moved and this PR now conflicts; re-entering the loop to rebase and re-verify with a reset conflict-fix budget.


mariushelf:
Rebase conflict resolved by conflict-fixer; fresh re-review follows. 

tests/api/conftest.py
```diff
@@ -337,6 +338,16 @@ def test_app(fake_orchestrator, fake_auth_orchestrator):
...
```
mariushelf, line 357: **minor** — this override is the right call for keeping the dozen route tests unmodified (AC 7 demands exactly that), but it does mean the *real* provider path is never exercised at route level: `get_summarize_service` is short-circuited, and every summarize route test still drives `FakeOrchestrator`. So after this PR nothing in the API tier would notice if `get_summarize_service` read the wrong `app.state` attribute — the only things standing between that and a 500 in production are `test_lifespan.py::test_summarize_service_is_published_on_app_state` and the e2e tier.

That's a real safety net (and the integration job is green on this SHA), so I'm not asking for a change here. Flagging it because the comment above explains *why the override exists* but not *what stops being covered because of it* — one sentence pointing at the lifespan test as the compensating check would keep the next reader from assuming the route tests cover the wiring.

src/qfa/services/summarize.py
```diff
@@ -0,0 +1,322 @@
...
+from qfa.domain.models import (
+    AggregateSummaryResultModel,
+    FeedbackRecordSummaryModel,
+    SingleSummaryRequestModel,
+    SummaryRequestModel,
+    SummaryResultModel,
+)
+from qfa.domain.ports import AnonymizationPort, LLMPort
+from qfa.services.llm_call_executor import LLMCallExecutor
+from qfa.services.orchestrator import _JUDGE_USER_MESSAGE, _hyperlink_form_references
```
mariushelf: **nit** — the `services.summarize → services.orchestrator` import is spec-sanctioned (the ticket says helpers still used by a retained use case "stay put"), and I verified both really do have other callers: `_hyperlink_form_references` at `orchestrator.py:415` and `:787`, `_JUDGE_USER_MESSAGE` at `:431` and `:884`. The module docstring flags it as temporary and names #267 as the rehoming ticket, which is what the AC asked for.

Worth noting for the human reviewer only: this points the *new* service at the *shrinking* god class across two private names, so every subsequent extraction in epic #112 inherits the same pattern and the last one to leave has to unpick all of them at once. If a `qfa.services.prompts`-style home shows up cheaply before then, moving these two early costs less than moving five later.

src/qfa/services/summarize.py
```diff
@@ -0,0 +1,322 @@
...
+
+        timeout = self._executor.check_deadline_and_get_timeout(deadline)
+        response = await self._llm.complete(
+            system_message=system_message,
+            user_message=anonymized_user_message,
+            tenant_id=request.tenant_id,
+            response_model=AggregateSummaryResultModel,
+            timeout=timeout,
+        )
+        total_cost = response.cost
```
mariushelf: **nit** — `total_cost` is assigned here and augmented at line 230, then never read. This is carried over verbatim from `Orchestrator.summarize_bulk` (I diffed the moved bodies against `origin/main` — they are byte-identical apart from the receiver), so it is *correct* for a behaviour-preserving refactor and I would not want it changed in this PR.

Calling it out because a move is the moment dead code becomes visible: in a 322-line single-purpose module it now reads like an accounting step that matters, and `test_orchestrator_judge_routing.py:518` has a docstring claiming `summarize_bulk` "is the one path that adds two call costs together in the service itself" — which describes this variable, and which the test does not actually assert (it asserts call counts). Fair game for a follow-up, not for this diff.

docs/architecture/03-components.md
```diff
@@ -95,24 +96,34 @@ recursing when a chunk or the combined partials overflow the token budget. See
...
+ports its own use cases need — `SummarizeService` asks for no embedder and no
+token ceiling, and says so by omission. Every service gets its own dependency
+provider in `qfa.api.dependencies` (`get_orchestrator`,
+`get_summarize_service`, …), so a route handler's signature names the one
+service it uses.
 
-Each method is pure use-case logic — no scope or correlation plumbing. `call_scope` is entered by a FastAPI dependency declared on the route (`Depends(call_scope_for(Operation.X))`), so by the time an orchestrator method runs `current_call_context` is already set. See [Cross-cutting concerns](04-crosscutting.md) for the full picture.
+Each method is pure use-case logic — no scope or correlation plumbing. `call_scope` is entered by a FastAPI dependency declared on the route (`Depends(call_scope_for(Operation.X))`), so by the time a use-case method runs `current_call_context` is already set. See [Cross-cutting concerns](04-crosscutting.md) for the full picture.
 
 ### Extracted use-case services
```
mariushelf, line 124: **nit** — `summarize`, `summarize_bulk` and `detect_sensitive_content` are now described twice on this page: once in the endpoint table above (which gained a "Service" column in this PR) and again in this "Extracted use-case services" table. Both are accurate today, and the duplication is arguably deliberate — the upper table is the endpoint map, this one the extraction ledger — but it means each remaining extraction in epic #112 has to touch two tables in the same file, and the failure mode is that one of them quietly keeps saying `Orchestrator`.

A "see below" pointer instead of repeating the description would leave one source of truth per fact. No action needed for this PR.

mariushelf (commented):
## Review — intent vs. behaviour (fresh context, post-rebase)

**Verdict: approve.** This is a behaviour-preserving extraction that does what #264 asked. I reviewed the rebased diff (`origin/main...HEAD`, merge base `06209fd`) as a first pass rather than trusting the pre-rebase reviews, and re-verified every acceptance criterion against the code.

### Acceptance criteria

| Criterion | Status | Evidence |
|---|---|---|
| `SummarizeService` in `services/summarize.py`, **no base class**, `LLMCallExecutor` ctor dep | met | `summarize.py:129` `class SummarizeService:`; `executor: LLMCallExecutor` is a required (non-default) parameter at `:162` |
| `summarize_bulk` / `summarize` moved **and deleted** from `Orchestrator` | met | −150 lines of method body from `orchestrator.py`; the six now-unused `qfa.domain.models` imports and both envelope helpers dropped with them |
| Shared helpers stay put; PR body says which and why | met | `_hyperlink_form_references` (live callers `orchestrator.py:415`, `:787`) and `_JUDGE_USER_MESSAGE` (`:431`, `:884`) stay and are imported; `_DEFAULT_*_PROMPT`, `_JUDGE_PROMPT`, `_build_judge_system_message`, `_parse_judge_quality_score` moved. One stale bullet in the PR body — see below |
| `get_summarize_service` in `api/dependencies.py` | met | `dependencies.py:57` |
| Both handlers annotate `SummarizeService`, no `Orchestrator` | met | `routes.py:268` and `:343` — parameter, annotation, docstring and call site all repointed |
| `composition.py` constructs it over the **one** `LLMCallExecutor` | met | single `executor = ...` at `composition.py:303` handed to all three services; asserted by identity (`is`), not type, in `test_services_share_one_executor` / `test_services_share_one_anonymizer` |
| Tests moved to `tests/services/test_summarize.py` over the real executor; nothing weakened | met | mechanically verified — see below |
| Existing e2e/API tests for both endpoints pass **unmodified** | met | `tests/api/test_routes.py` and `tests/e2e/` have zero diff, and the `integration` job (`make test-integration`) is **green on this exact SHA** `2163f89` |
| `make test` / `make lint` incl. all three import-linter contracts | met | reproduced locally: **625 passed, 42 deselected**; ruff + `ty` clean; **3 contracts kept, 0 broken** |

### How I checked the "nothing weakened" claim

I did not take it on faith. I extracted the 14 moved test bodies from `origin/main:tests/services/test_orchestrator.py` and diffed each against its counterpart in the new module. Every one differs **only** by swapping the six-line `Orchestrator(...)` construction for `_build_service(...)` and the receiver name — no assertion, no `pytest.raises` match, no fixture value, no `==` operand changed. The three the ticket names are all present and intact: `test_output_language_instructs_aggregate_system_message` / `test_no_language_directive_when_output_language_unset`, `test_hyperlinks_form_reference_in_summary` / `test_no_hyperlink_without_base_url`, and both `*_forbids_trailing_questions`. `_build_service` builds the **real** `LLMCallExecutor` over `FakeLLMPort` / `FakeAnonymizer` — no fake executor — and threads `max_total_tokens` through, so `TestTokenLimit` still asserts a large payload *is* forwarded.

The move also had two same-named-test traps in it, and both were navigated correctly. `main` had two `test_no_hyperlink_without_base_url` (the summarize one at `test_orchestrator.py:723`, the analyze one at `:1195`) and two `test_no_language_directive_when_output_language_unset` (`:663` summarize, `:598` analyze). In each pair the summarize copy moved and the analyze copy stayed — I diffed the bodies pairwise to confirm the right one went each way, since taking the wrong one would have silently deleted analyze-side coverage while every test name still looked accounted for.

The moved production code is byte-identical too: `diff` of `orchestrator.py:1137-1286` on `main` against `summarize.py:174-322` is empty modulo indentation.

### Post-rebase check

The rebase onto #263 (`SensitivityService`) is what I focused hardest on, since the pre-rebase reviews cannot speak to it. It is clean: `build_service_graph` returns a three-field graph, all three services share one executor *and* one anonymiser (both asserted by identity), the lifespan publishes all three slots, and `test_sensitivity.py` / `test_llm_call_executor.py` come through untouched. `deanonymize_json` now lives on the executor and `summarize.py` correctly reaches it through `self._executor` rather than importing the old orchestrator helper.

### Notes for the human reviewer

- **The PR body has one stale bullet.** It lists `_json_escape_mapping` among the helpers that "stayed in `qfa.services.orchestrator` ... and are imported from there (still used by `detect_sensitive_content`)". After the rebase neither half is true: `_json_escape_mapping` lives at `src/qfa/services/llm_call_executor.py:41` (moved by #262), `summarize.py` does not import it, and `detect_sensitive_content` is on `SensitivityService`. The code is right; only the description is out of date. The other two bullets in that list I verified against live call sites and they hold.
- **`docs/operations/settings-reference.md:23-24`** still reads "It applies to four call sites: the `analyze` judge, the hierarchical leaf judges, and the judges in `summarize` and `summarize_aggregate`", as though all four sat on the orchestrator. This PR updated the parallel sentence in `docs/architecture/03-components.md` but not this one. It isn't *wrong* — the sentence names methods, and `JUDGE_LLM_*` still routes exactly those four — so it's a nit. (`summarize_aggregate` for the method now called `summarize_bulk` is a repo-wide inconsistency that predates this PR.)
- **`build_orchestrator` now builds a `SummarizeService` it discards** for scripts and notebooks. Free in practice (the constructor does no I/O) and it buys a single wiring path, which `test_build_orchestrator_returns_the_graphs_orchestrator` pins. Noting it so it isn't mistaken for an oversight.
- Four further nits are inline on the diff.

### Could not verify

- Whether the cross-service judge-routing tests belong in `tests/services/test_orchestrator_judge_routing.py` or should follow the use case into `test_summarize.py` is a taste call. The PR argues for keeping them together — "splitting them per service is exactly how one call site quietly stops being checked" — which I find persuasive, but the module name now undersells its scope, since two of the four call sites it covers are no longer on the orchestrator. A human should settle whether the file wants renaming.

mariushelf:
Review iteration 2: **approve** — Behaviour-preserving extraction that does what #264 asked. Reviewed the rebased diff (`origin/main...HEAD`, merge base `06209fd`) as a fresh pass. `SummarizeService` is a plain class with no base class taking a required `LLMCallExecutor`; both methods are deleted from `Orchestrator` along with their now-unused model/prompt imports; the two helpers still used by retained use cases (`_hyperlink_form_references` at orchestrator.py:415/:787, `_JUDGE_USER_MESSAGE` at :431/:884) stay put and are imported, with the temporary coupling flagged in both docstrings and tracked to #267. `get_summarize_service` exists; both handlers annotate `SummarizeService`; `build_service_graph` builds one executor and one anonymiser and hands the same instances to all three services, asserted by identity rather than type. Verified the moved production code is byte-identical (diff of orchestrator.py:1137-1286 on main vs summarize.py:174-322 is empty modulo indentation), and mechanically diffed all 14 moved test bodies against origin/main: each differs only by swapping the `Orchestrator(...)` construction for `_build_service(...)` and the receiver name — no assertion, `pytest.raises` match, or fixture value changed. Both same-named-test traps in the source file (two `test_no_hyperlink_without_base_url`, two `test_no_language_directive_when_output_language_unset`) were navigated correctly: the summarize copy moved, the analyze copy stayed, confirmed by pairwise body diff. `_build_service` uses the real `LLMCallExecutor` over `FakeLLMPort`/`FakeAnonymizer`. Reproduced `make test` (625 passed, 42 deselected) and `make lint` (ruff + `ty` clean, 3 import-linter contracts kept, 0 broken); `tests/api/test_routes.py` and `tests/e2e/` have zero diff and the CI `integration` job (`make test-integration`) is green on this exact SHA 2163f89, which closes the earlier rounds' e2e gap. Post-rebase wiring onto #263 is clean — `deanonymize_json` now lives on the executor and summarize.py reaches it through `self._executor`. Five findings, none gating: one minor (the tests/api/conftest.py override means no route-level test exercises the real provider) and four nits, one of which is a stale bullet in the PR description itself.


mariushelf:
Could not verify: Whether the cross-service judge-routing tests belong in `tests/services/test_orchestrator_judge_routing.py` or should follow the use case into `tests/services/test_summarize.py` is a taste call for the human reviewer. The PR's argument for keeping them together ('splitting them per service is exactly how one call site quietly stops being checked') is persuasive, but the module name now undersells its scope, since two of the four call sites it covers are no longer on the orchestrator. Whether the file wants renaming is a human judgement, not something the diff can settle.


mariushelf:
`main` moved and this PR now conflicts; re-entering the loop to rebase and re-verify with a reset conflict-fix budget.


mariushelf:
Rebase conflict resolved by conflict-fixer; fresh re-review follows. 

mariushelf (commented):
## Review — intent vs. behaviour (fresh context, post-rebase #3)

**Verdict: request changes.** The extraction itself is correct and faithful — I re-verified every acceptance criterion from scratch rather than trusting the two earlier approvals, and the design half of this PR holds up. But **the branch is red on this SHA**, and not on something cosmetic: `make test` fails, `make lint` fails, and a retained `assign_codes` assertion has stopped being checked. One line fixes it.

### The blocker

`tests/services/test_orchestrator.py` lost its `AnalysisError` import along with the two summarize judge tests that used it — but a *third*, retained use survives at `:1257`, in `TestAssignCodesOneShot::test_judge_score_out_of_range_raises_analysis_error`. That test belongs to `assign_codes` and does not move with this PR.

```
$ make test
===== 1 failed, 641 passed, 42 deselected, 1 warning in 120.86s =====
FAILED tests/services/test_orchestrator.py::TestAssignCodesOneShot::test_judge_score_out_of_range_raises_analysis_error
E       NameError: name 'AnalysisError' is not defined

$ make lint
F821 Undefined name `AnalysisError`  --> tests/services/test_orchestrator.py:1257:28
make: *** [Makefile:33: ruff] Error 1

$ uv run ty check
error[unresolved-reference]: Name `AnalysisError` used when not defined
```

CI confirms it independently: `lint (3.12)`, `lint (3.13)`, `test (3.12)` and `test (3.13)` are all failing on `f234060`. Because `ruff` is the first step of `make lint`, `ty` and `lint-imports` never run in CI — I ran both separately, and `lint-imports` reports **3 contracts kept, 0 broken**, so the layer rules are fine and the breakage is confined to this one import.

The cost is more than a red build. `pytest.raises(AnalysisError, ...)` raises `NameError` while evaluating its own argument, so the context manager is never armed and the guarantee it encodes — an out-of-range judge score in `assign_codes` is a domain error, not a generic LLM failure — is no longer verified. Restoring `AnalysisError` to the `from qfa.domain.errors import (...)` block is the whole fix.

This reads as fallout from the third rebase, not from the original extraction: both earlier rounds reproduced a green `make test` (620 and 625 passed), and the `assign_codes` judge test that trips it landed on `main` in between. Worth a glance at whether the conflict resolution touched anything else in that file — I checked the rest of the pruned imports and they are all genuinely unused, so this is the only casualty I found.

### Acceptance criteria

| Criterion | Status | Evidence |
|---|---|---|
| `SummarizeService` in `services/summarize.py`, **no base class**, `LLMCallExecutor` ctor dep | met | `summarize.py:129` `class SummarizeService:`; `executor: LLMCallExecutor` is a required (non-default) parameter at `:162` |
| `summarize_bulk` / `summarize` moved **and deleted** from `Orchestrator` | met | −150 lines of method body; the six now-unused `qfa.domain.models` imports and both envelope helpers dropped with them |
| Shared helpers stay put; PR body says which and why | met | `_hyperlink_form_references` (live callers `orchestrator.py:415`, `:787`) and `_JUDGE_USER_MESSAGE` (`:431`, `:884`) stay and are imported; `_DEFAULT_*_PROMPT`, `_JUDGE_PROMPT`, `_build_judge_system_message`, `_parse_judge_quality_score` moved. PR body states the split — with one stale bullet, see below |
| `get_summarize_service` in `api/dependencies.py` | met | `dependencies.py:57` |
| Both handlers annotate `SummarizeService`, no `Orchestrator` | met | `routes.py:268`, `:343` — parameter, annotation, docstring and call site all repointed |
| `composition.py` constructs it over the **one** `LLMCallExecutor` | met | single `executor = ...` at `:303` handed to all three services; asserted by identity (`is`), not type, in `test_services_share_one_executor` / `test_services_share_one_anonymizer` |
| Tests moved to `tests/services/test_summarize.py` over the real executor; nothing weakened | met | mechanically verified — see below |
| Existing e2e/API tests for both endpoints pass **unmodified** | met | `tests/api/test_routes.py` and `tests/e2e/` have zero diff; `tests/api` is 184 passed locally, and the **`integration` CI job passes on this exact SHA** |
| `make test` / `make lint` pass, incl. all three `import-linter` contracts | **not met** | `make test`: 1 failed / 641 passed. `make lint`: ruff F821 + `ty` unresolved-reference. `lint-imports` alone: 3 kept, 0 broken |

Note the blocker sits in AC 9, not AC 7: what broke is a *retained* `assign_codes` test in the file the move left behind, not one of the moved summarize tests. Those came over clean.

### How I checked "no assertion weakened"

I did not take it on faith. I parsed both modules with `ast`, extracted all 14 moved test bodies from `origin/main:tests/services/test_orchestrator.py`, and compared each against its counterpart in the new module. All 14 are present and every assertion, `pytest.raises` match and fixture value matches after normalising the receiver name — nothing missing, nothing softened. The three the ticket names are intact: `test_output_language_instructs_aggregate_system_message` / `test_no_language_directive_when_output_language_unset`, `test_hyperlinks_form_reference_in_summary` / `test_no_hyperlink_without_base_url`, and both `*_forbids_trailing_questions`.

`_build_service` builds the **real** `LLMCallExecutor` over `FakeLLMPort` / `FakeAnonymizer` — no fake executor — and threads `max_total_tokens` through so `TestTokenLimit` still asserts a large payload *is* forwarded.

The move contained two same-named-test traps and both were navigated correctly. `main` had two `test_no_hyperlink_without_base_url` and two `test_no_language_directive_when_output_language_unset`; in each pair the summarize copy moved and the analyze copy stayed (now at `test_orchestrator.py:771` and `:365`). Taking the wrong one would have deleted analyze-side coverage while every test name still looked accounted for.

The moved production code is byte-identical: a whitespace-normalised diff of `summarize_bulk` + `summarize` against `origin/main:src/qfa/services/orchestrator.py` is empty across all 130 lines.

### What else I verified

- **Post-rebase wiring onto #263 is clean.** `build_service_graph` returns a three-field graph; all three services share one executor *and* one anonymiser (both asserted by identity); the lifespan publishes all three slots; `deanonymize_json` now lives on the executor and `summarize.py` reaches it through `self._executor` rather than the old orchestrator helper.
- **Docs are thorough and accurate**, including the `#the-orchestrator` → `#the-application-services` anchor rename and its one inbound reference in `implementing-a-new-endpoint.md:103`. I grepped every `03-components.md#` link; none dangle. `make docs` was not re-run here.
- **No stale `build_services` reference** survives anywhere in the tree.

### Notes for the human reviewer

- **The PR body still has one stale bullet**, unchanged from the last round: it lists `_json_escape_mapping` among helpers that \"stayed in `qfa.services.orchestrator` … and are imported from there (still used by `detect_sensitive_content`).\" Neither half is true post-rebase — it lives at `src/qfa/services/llm_call_executor.py:41`, `summarize.py` does not import it, and `detect_sensitive_content` is on `SensitivityService`. The code is right; only the description is out of date. The other two bullets I re-verified against live call sites and they hold.
- **`docs/operations/settings-reference.md:23-24`** still reads \"It applies to four call sites: the `analyze` judge, the hierarchical leaf judges, and the judges in `summarize` and `summarize_aggregate`,\" as though all four sat on the orchestrator. The parallel sentence in `docs/architecture/03-components.md` was updated; this one was not. It isn't *wrong* — it names methods, and `JUDGE_LLM_*` still routes exactly those four — so it stays a nit. (`summarize_aggregate` for the method now called `summarize_bulk` is a repo-wide inconsistency predating this PR.)
- **`build_orchestrator` now builds a `SummarizeService` it discards** for scripts and notebooks. Free in practice (the constructor does no I/O) and it buys a single wiring path, which `test_build_orchestrator_returns_the_graphs_orchestrator` pins. Noting it so it isn't mistaken for an oversight.
- Five further nits/minors are inline on the diff.

### Could not verify

- Whether the cross-service judge-routing tests belong in `tests/services/test_orchestrator_judge_routing.py` or should follow the use case into `test_summarize.py` is a taste call. The PR argues for keeping them together — \"splitting them per service is exactly how one call site quietly stops being checked\" — which I find persuasive, but the module name now undersells its scope, since two of the four call sites it covers are no longer on the orchestrator. Whether the file wants renaming is a human judgement.
- `make docs` was not re-run in this round; the `build-docs` CI job passes on this SHA.

tests/services/test_orchestrator.py
```diff
@@ -6,24 +6,18 @@
 import pytest
 
 from qfa.domain.errors import (
-    AnalysisError,
     LLMError,
```
mariushelf: **Blocker — `AnalysisError` is still used in this file, so `make test` and `make lint` are both red.**

The move dropped `AnalysisError` from this import block along with the two summarize judge tests that used it (`origin/main` lines 430 and 451). But a *third*, retained use survives at `tests/services/test_orchestrator.py:1257` — `TestAssignCodesOneShot::test_judge_score_out_of_range_raises_analysis_error`, which belongs to `assign_codes` and does not move with this PR (it was line 1683 on `origin/main`).

Reproduced on this SHA:

```
$ uv run pytest tests/services/test_orchestrator.py -k out_of_range
>       with pytest.raises(AnalysisError, match=r"outside 0\.0-1\.0"):
E       NameError: name 'AnalysisError' is not defined
tests/services/test_orchestrator.py:1257: NameError

$ make test
===== 1 failed, 641 passed, 42 deselected, 1 warning in 120.86s =====
make: *** [Makefile:18: test-unit-tests] Error 1

$ make lint
F821 Undefined name `AnalysisError`
    --> tests/services/test_orchestrator.py:1257:28
make: *** [Makefile:33: ruff] Error 1

$ uv run ty check
error[unresolved-reference]: Name `AnalysisError` used when not defined
    --> tests/services/test_orchestrator.py:1257:28
```

CI agrees: `lint (3.12)`, `lint (3.13)`, `test (3.12)` and `test (3.13)` are all failing on `f234060`. (`integration` passes, and `lint-imports` reports 3 contracts kept / 0 broken when run past the ruff failure — the breakage is confined to this one import.)

The cost is not only the red build: an `assign_codes` guarantee that a *retained* use case still owns — an out-of-range judge score is an `AnalysisError`, not a generic LLM failure — now raises `NameError` before its `pytest.raises` is ever armed, so that assertion silently stops being checked.

Fix is one line — restore the name:

```python
from qfa.domain.errors import (
    AnalysisError,
    LLMError,
    LLMResponseParseError,
)
```

This looks like fallout from the third rebase rather than the original extraction: the two earlier review rounds both reproduced a green `make test` (620 and 625 passed), and the `assign_codes` judge test that trips it arrived on `main` in between.

tests/api/conftest.py
mariushelf, line 357: **Minor — after this override, no route-level test exercises `SummarizeService` or the real provider.**

The override is well-reasoned and the comment above it earns its keep: a dozen tests in `test_routes.py` swap `state.orchestrator` for an error-raising fake *after* this fixture runs, and reading `state.orchestrator` per request is what keeps them working without edits. That is genuinely why AC 8 ("e2e/API tests pass **unmodified**") holds.

The trade-off is that `get_summarize_service` and the `app.state.summarize_service` slot are now bypassed in every API test, so the provider→state contract is only pinned by `test_lifespan.py::test_summarize_service_is_published_on_app_state` and not by anything that actually issues a request. That is thinner than what the sensitivity route got, and it is worth a human's call on whether one route test resolving the real provider over a `SummarizeService` built on fakes is worth adding.

Not gating: the lifespan test does assert publication over the real composition path, and the `integration` CI job (which boots the real `create_app`) passes on this SHA.

src/qfa/services/summarize.py
mariushelf: **Nit — underscore-prefixed names imported across module boundaries.**

This is sanctioned: the spec says a helper still needed by a retained use case "stays put," both `_hyperlink_form_references` (live at `orchestrator.py:415` and `:787`) and `_JUDGE_USER_MESSAGE` (`:431`, `:884`) qualify, the module docstring flags the coupling as temporary, and #267 is named as the rehoming ticket. I verified all four call sites.

Still worth a note for the human reviewer: the direction of the arrow is now `summarize → orchestrator`, which is the opposite of what the decomposition is heading toward, and a leading underscore normally means "do not import this from elsewhere." `qfa.services.prompts` already exists and is where `build_feedback_record_envelope` and friends live, so it is the natural home for both — the shape works today and only reads oddly while #267 is open.

src/qfa/services/summarize.py
mariushelf: **Nit — `total_cost` is accumulated and never read.**

`total_cost` is set here and incremented at `:230`, then nothing consumes it: the method returns a `model_copy` of the validated result and cost accounting happens in `TrackingLLMAdapter`. Dead on `origin/main` too (`orchestrator.py:1181` / `:1195`), so this is faithfully carried over rather than introduced — and the whitespace-normalised diff of both moved methods against `main` is empty, which is exactly what a behaviour-preserving move should look like.

Mentioning it only because a new module is the cheapest moment to drop it, and because `test_summarize_bulk_adds_both_costs` in `test_orchestrator_judge_routing.py` reads as though it pins this variable when what it actually asserts is one call on each client.

src/qfa/api/composition.py
```diff
@@ -195,14 +206,14 @@ def register_custom_model_prices() -> None:
         )
 
 
-def build_services(
+def build_service_graph(
```
mariushelf: **Nit — the `build_services` → `build_service_graph` rename is now churn the ticket didn't ask for.**

When this branch was first written the function did not exist, so introducing it under this name was free. After the rebase onto #263, `main` already ships `build_services(...) -> ServiceGraph`, and this PR renames it plus every reference (`app.py`, `test_composition.py`, `test_lifespan.py`, three docs pages, two service docstrings). All call sites are updated — I grepped and there is no surviving `build_services` reference anywhere in the tree — so nothing is broken; it just widens the diff of an extraction PR with a rename that reviewers of #263 will have to re-read.

A human may well prefer the new name (it does read better next to `build_orchestrator`). Flagging it as scope, not as a defect.

tests/api/test_lifespan.py
```diff
@@ -166,7 +187,7 @@ async def test_judge_model_alone_is_a_valid_startup_configuration(
 
 
 @pytest.mark.asyncio
-async def test_every_service_is_published_on_app_state(app_env: None):
+async def test_sensitivity_service_is_published_on_app_state(app_env: None):
     """Each extracted use-case service gets its own ``app.state`` slot.
```
mariushelf: **Nit — docstring now overclaims after the rename.**

"Each extracted use-case service gets its own `app.state` slot" was accurate while this test was `test_every_service_is_published_on_app_state` and covered the whole set. Now that it is split per service, this body asserts only `sensitivity_service`, so the summary sentence promises more than the test checks — a reader adding a fourth service could reasonably conclude this test would catch a missed publication, when in fact it needs its own sibling (as `test_summarize_service_is_published_on_app_state` correctly is).

The split itself is the right call: per-service tests name which slot is missing when one fails.

mariushelf:
Review iteration 3: **request_changes** — Reviewed the rebased diff (`origin/main...HEAD`, merge base `5fba9bb`, SHA `f234060`) as a fresh first pass. The extraction itself is correct and faithful, but the branch is red and one gating defect must be fixed before a human sees it.\n\nBlocker: `tests/services/test_orchestrator.py` lost its `AnalysisError` import along with the two moved summarize judge tests, but a retained `assign_codes` test at `:1257` (`TestAssignCodesOneShot::test_judge_score_out_of_range_raises_analysis_error`) still uses it. Reproduced locally: `make test` = 1 failed / 641 passed; `make lint` = ruff F821 plus `ty` unresolved-reference. CI confirms independently — `lint (3.12)`, `lint (3.13)`, `test (3.12)` and `test (3.13)` all fail on this exact SHA. Because ruff is the first step of `make lint`, `ty` and `lint-imports` never run in CI; I ran both separately and `lint-imports` reports 3 contracts kept / 0 broken, so the layer rules are fine and the breakage is confined to this one import. Beyond the red build, `pytest.raises(AnalysisError, ...)` raises `NameError` while evaluating its own argument, so that `assign_codes` guarantee has silently stopped being checked. This looks like fallout from the third rebase — both earlier rounds reproduced a green `make test` (620 and 625 passed) and the `assign_codes` judge test that trips it landed on `main` in between. One-line fix: restore `AnalysisError` to the import block.\n\nEverything else holds. `SummarizeService` is a plain class with no base class taking a required `LLMCallExecutor`; both methods are deleted from `Orchestrator` along with their now-unused model and envelope imports; the two helpers with live retained callers (`_hyperlink_form_references` at `orchestrator.py:415`/`:787`, `_JUDGE_USER_MESSAGE` at `:431`/`:884`) stay put and are imported, with the coupling tracked to #267; `get_summarize_service` exists at `dependencies.py:57`; both handlers annotate `SummarizeService` at `routes.py:268`/`:343`; `build_service_graph` builds one executor and one anonymiser and hands the same instances to all three services, asserted by identity rather than type. The moved production code is byte-identical — a whitespace-normalised diff of `summarize_bulk` + `summarize` against `origin/main` is empty across all 130 lines. I parsed both test modules with `ast` and compared all 14 moved test bodies against `origin/main`: all present, every assertion, `pytest.raises` match and fixture value intact after normalising the receiver name, including the ticket's three named cases. Both same-named-test traps were navigated correctly (the summarize copy moved, the analyze copy stayed at `:771` and `:365`). `_build_service` uses the real `LLMCallExecutor` over `FakeLLMPort`/`FakeAnonymizer`. `tests/api/test_routes.py` and `tests/e2e/` have zero diff, `tests/api` is 184 passed locally, and the `integration` CI job passes on this SHA, which closes the earlier rounds' e2e gap. Docs are thorough, including the anchor rename and its one inbound reference; no stale `build_services` reference survives.\n\nSix findings: one blocker, one minor (the `tests/api/conftest.py` override means no route-level test exercises the real provider), and four nits — including a stale `_json_escape_mapping` bullet in the PR description, unchanged from the previous round.


mariushelf:
**Escalated:** Review-iteration cap (3) hit; still contested:
- [blocker] tests/services/test_orchestrator.py: `AnalysisError` was dropped from the `from qfa.domain.errors import (...)` block along with the two moved summarize judge tests, but a third, retained use survives at `tests/services/test_orchestrator.py:1257` in `TestAssignCodesOneShot::test_judge_score_out_of_range_raises_analysis_error` (an `assign_codes` test that does not move with this PR). Result: `make test` fails (1 failed, 641 passed), `make lint` fails (ruff F821 and `ty` unresolved-reference), and CI `lint (3.12)`, `lint (3.13)`, `test (3.12)`, `test (3.13)` are all red on f234060. Beyond the red build, `pytest.raises(AnalysisError, ...)` raises `NameError` while evaluating its own argument, so the context manager is never armed and the assertion that an out-of-range judge score in `assign_codes` is an `AnalysisError` rather than a generic LLM failure has silently stopped being checked. Appears to be fallout from the third rebase: both earlier rounds reproduced a green `make test` (620 and 625 passed) and the `assign_codes` judge test that trips it landed on `main` in between. `lint-imports` run separately reports 3 contracts kept, 0 broken, so the layer rules are unaffected and the breakage is confined to this one import.


mariushelf:
/loop fix merge conflicts. Things already merged to main as part of the previous 4 PRs should lead, e.g. when establishing function and method names. This PR should NOT rename functions or duplicate concepts.

mariushelf:
`/loop fix` received — re-entering the loop with reset counters.


mariushelf:
Stage conflict_fix retried (crash: error_max_turns). 

mariushelf:
Rebase conflict resolved by conflict-fixer; fresh re-review follows. 

mariushelf:
**Escalated:** Fixer edited test file(s) no gating finding sanctions: tests/api/test_composition.py, tests/services/test_analyze.py, tests/services/test_coding.py, tests/services/test_llm_call_executor.py, tests/services/test_orchestrator_judge_routing.py, tests/services/test_sensitivity.py


mariushelf:
/loop review review this. focus on the changes since the last review, with on focus on whether the previously merged PRs AND the work of this PR survive. Check that neither PR duplicates a feature/concept (e.g. in composition.py).

mariushelf:
`/loop review` received — re-reviewing the PR head as it stands; the reviewer's verdict is the end of it, so post `/loop fix` if you want a fixer round on the findings. A branch that does not yet contain the base is still brought up to date first, and the conflict-fixer runs if that conflicts.


tests/api/conftest.py
```diff
@@ -339,6 +340,22 @@ def test_app(fake_orchestrator, fake_auth_orchestrator):
...
+    # error-raising fake *after* this fixture ran, and they must keep
+    # driving the summarize endpoints without knowing which service class
+    # is behind them.
+    #
+    # The cost: no test in this package resolves the real provider, so the
+    # provider/``app.state`` contract is pinned only by
+    # ``test_lifespan.py::test_every_service_is_published_on_app_state``,
+    # which drives the real composition root. Check there, not here, when
+    # changing which slot ``get_summarize_service`` reads.
+    app.dependency_overrides[get_summarize_service] = lambda: app.state.orchestrator
```
mariushelf, line 357: **minor — this is a second wiring mechanism for a concept the three merged sibling PRs already settled.**

The three services extracted before this one are handed to the route tier the same way, three lines above: `app.state.sensitivity_service = fake_orchestrator`, `coding_service`, `analyze_service`. Summarize is the only one resolved through `dependency_overrides`. Two shapes now mean the same thing in one fixture, and `#267` (or whoever adds the fifth service) has to notice which of the two to copy.

I don't think it's wrong, and the comment above it is unusually honest about the trade-off. The reason it exists is that AC 7 demands the route tests pass **unmodified**, and `test_routes.py:788-846` (`test_summary_413_feedback_too_large` through `test_summary_500_unexpected_exception`) inject their error-raising fake by reassigning `test_app.state.orchestrator` *after* this fixture ran — a static `app.state.summarize_service = fake_orchestrator` would leave those five tests driving the original fake and silently stop asserting the error mapping. The prior PRs solved the same problem by editing the route tests (`test_app.state.analyze_service = FakeOrchestrator(error=...)` at `:715-777`, `sensitivity_service` at `:869`), which this ticket's AC forbids.

So: the divergence is the ticket's doing, not the author's, and it is correctly reasoned. Flagging it for the human because it is the one place where this PR's shape does not match what already landed on `main`, and because the consequence noted in the comment is real — no API-tier test resolves the real `get_summarize_service`, so a provider reading the wrong `app.state` slot is caught only by `test_lifespan.py::test_every_service_is_published_on_app_state` and the e2e tier. Cheapest top-up, if a human wants one: leave the override but point it at `MagicMock(spec=SummarizeService)` in one test, so the fake is conformance-checked against the real class at least once.

src/qfa/api/dependencies.py
```diff
@@ -95,6 +96,25 @@ def get_analyze_service(request: Request) -> AnalyzeService:
     return request.app.state.analyze_service
 
 
+def get_summarize_service(request: Request) -> SummarizeService:
```
mariushelf, line 99: **nit — `get_orchestrator` (`:23`) is dead as of this commit.**

With both summarize handlers repointed here, nothing in `src/`, `tests/` or `docs/` calls `get_orchestrator` any more — it was the last route provider still wired to `Orchestrator`. `03-components.md:120` correctly drops it from the provider list in this PR, so the docs and the code now disagree about whether it exists.

Leaving it is defensible (`#267` deletes the class and this function with it, and removing it here widens the diff), but it is worth the human knowing that it is now an unreferenced public function rather than something the routes still need.

docs/architecture/03-components.md
```diff
@@ -166,7 +175,7 @@ Tests construct the real executor over the existing `FakeLLMPort` / `FakeAnonymi
...
 
-The lifespan then attaches each service (`app.state.orchestrator`, `app.state.sensitivity_service`, `app.state.coding_service`, `app.state.analyze_service`), the API keys, and the usage repository to `app.state` for the request lifecycle to read. One `app.state` slot per service is what lets each route's provider inject only the use case it needs.
+The lifespan then attaches each service (`app.state.orchestrator`, `app.state.sensitivity_service`, `app.state.coding_service`, `app.state.analyze_service`, `app.state.summarize_service`), the API keys, and the usage repository to `app.state` for the request lifecycle to read. One `app.state` slot per service is what lets each route's provider inject only the use case it needs.
```
mariushelf, line 178: **minor — the paragraph immediately below (`:180`) is now inaccurate, and this PR is what made it so.**

> The split exists so callers outside the API server — scripts, notebooks, ad-hoc evaluation harnesses — can construct the services over a plain LLM client with a single call ({py:func}`~qfa.api.composition.build_orchestrator` and {py:func}`~qfa.api.composition.build_analyze_service` are the narrow wrappers over `build_services` for exactly that case)

After this commit `build_orchestrator` returns an object with no use case on it, so it is no longer a way for a script to "construct the services" — a notebook author following this page gets an empty shell. The docstring in `composition.py:357` *was* updated to say exactly that ("Convenience wrapper for callers still written against the class — it holds no use case since #264"), so the code is right and only this page lags. The same sentence recurs at `:190` ("its single-service wrappers `build_orchestrator` and `build_analyze_service`").

One clause — naming `build_analyze_service` as the wrapper scripts want and `build_orchestrator` as vestigial — would close it. Non-blocking; `make docs` builds and no link dangles.

tests/services/test_analyze.py
```diff
@@ -27,10 +27,10 @@
 from qfa.services.prompts import JUDGE_UNAVAILABLE_EXPLANATION
 from qfa.settings import OrchestratorSettings
 
-# Reuse the doubles the orchestrator suite already ships rather than growing a
+# Reuse the doubles the summarize suite already ships rather than growing a
 # second, drifting pair (ADR-017) — same call the sensitivity and coding
 # suites make.
-from .test_orchestrator import FakeAnonymizer, FakeLLMPort
+from .test_summarize import FakeAnonymizer, FakeLLMPort
```
mariushelf, line 33: **nit — the shared test doubles now live in a use-case-specific module.**

`FakeLLMPort` / `FakeAnonymizer` are imported by four sibling suites (`test_analyze`, `test_coding`, `test_sensitivity`, `test_llm_call_executor`). They used to sit in `test_orchestrator.py`, which at least read as the generic module; after the wholesale rename they sit in `test_summarize.py`, so four suites now import their doubles from the summarisation tests specifically. Repointing them was the minimal correct move for this PR and the updated comments are accurate.

Worth noting the direction: this is the second time these imports have been rewritten across four files (`#266` did it once), and `#267` will have the same choice again. A `tests/services/fakes.py` or a `conftest.py` fixture would end the churn and stop the doubles from appearing to belong to whichever use case happened to move last. Not for this diff.

src/qfa/services/summarize.py
```diff
@@ -0,0 +1,320 @@
...
+    ``quality_score``.
+
+    Parameters
+    ----------
+    llm : LLMPort
+        The LLM provider adapter used for the summary generation calls.
+    anonymizer : AnonymizationPort
+        The anonymisation adapter used to redact PII before the LLM calls
+        and to restore it in the result.
+    executor : LLMCallExecutor
```
mariushelf, line 143: **nit — the `executor` docstring undersells what these two use cases take from it.**

"These two use cases consult it for the deadline→timeout derivation" — they also call `self._executor.deanonymize_json(...)` on the way out (`:234`, `:311`), which is the step that restores PII into the result. Given the class docstring's job is to say what the collaborator is *for*, the anonymisation round-trip is the more consequential of the two to name.

mariushelf (commented):
Review iteration 4 (post-conflict-fix): **approve** — reviewed `origin/main...HEAD` fresh (merge base `4ee1cce`, head `fffc997`, 20 files, +656/-521), with the focus the last comment asked for: does the work of the four already-merged PRs survive, does this PR's own work survive the rebase, and does anything get duplicated or renamed.

**All nine acceptance criteria are met.** The blocker from iteration 3 (`AnalysisError` dropped from `tests/services/test_orchestrator.py`'s import block while a retained `assign_codes` test still used it) is gone — the file that carried it no longer exists on this branch, and the `assign_codes` tests now live on `main` in `tests/services/test_coding.py`. Reproduced green locally: `make test` = **651 passed, 42 deselected**, `make lint` = all checks passed with **3 contracts kept, 0 broken**. Every CI check on this SHA is green, including `integration`, `test (3.12)`, `test (3.13)`, both `lint` jobs and `build-docs`.

### Nothing was renamed, nothing was duplicated

This was the specific ask, so I checked it directly rather than by inspection of the summary:

- **`build_services` keeps `main`'s name.** The `build_services` → `build_service_graph` rename I flagged as scope churn in an earlier round is gone; `composition.py` now has exactly the top-level surface `main` ships (`ServiceGraph`, `resolve_judge_llm_settings`, `build_embedder`, `register_custom_model_prices`, `build_services`, `build_orchestrator`, `build_analyze_service`) plus one new `summarize: SummarizeService` field on the dataclass. No sibling function, no second graph builder.
- **No helper is duplicated across services.** I grepped every judge/hyperlink symbol in `src/`. `hyperlink_form_references` has one definition (`qfa.services.record_links`), imported by both `analyze.py` and `summarize.py`; `JUDGE_USER_MESSAGE` has one definition (`qfa.services.prompts`), likewise. `summarize.py` defines nothing that already existed elsewhere, and — unlike every earlier round — it reaches into `qfa.services.orchestrator` for nothing at all. The `summarize → orchestrator` private-import coupling the previous reviews flagged has been fully unwound by the rebase.
- **Naming follows the established convention rather than adding a fourth spelling**: `ServiceGraph.summarize` / `app.state.summarize_service` / `get_summarize_service`, matching `sensitivity` / `coding` / `analyze`.

### The extraction itself

`SummarizeService` is a plain class, no base class, with `executor: LLMCallExecutor` a **required** constructor parameter (`summarize.py:158`) — no `= None` default, so nothing can construct one that quietly builds its own. Both methods are deleted from `Orchestrator` along with every import and module constant they alone used; what's left there is a constructor and nothing else.

The moved bodies are faithful. A whitespace-normalised diff of `summarize_bulk` + `summarize` against `origin/main:src/qfa/services/orchestrator.py` differs in exactly the places the PR body claims: the receiver, and the removal of the `total_cost` accumulator. That variable was assigned and incremented but never read on `main` either — cost is summed per client by `TrackingLLMAdapter` — so dropping it is behaviour-neutral, and it is disclosed in the PR description rather than slipped in.

Wiring: `build_services` constructs it over the *same* `LLMCallExecutor` and `PresidioAnonymizer` as every other service, asserted by **identity** (`services.summarize._executor is services.orchestrator._executor`, same for `_anonymizer`) rather than by type; the lifespan publishes `app.state.summarize_service`; `test_every_service_is_published_on_app_state` pins it over the real composition root; `test_summarize_service_gets_both_connections` pins that the judge client actually reaches the service, which is the failure that would otherwise silently move two judge call sites back onto the generation model. Both handlers annotate `SummarizeService` and no longer import `Orchestrator`.

### Tests: nothing weakened

`test_orchestrator.py` → `test_summarize.py` is a git-detected rename at 71% similarity, so I checked it test by test rather than trusting the header. All 8 classes and all 14 test functions on `main` are present with the same names. The removals are three genuinely-dead module helpers (`_make_analysis_result`, `_past_deadline`, the `orchestrator` fixture) — `_past_deadline` has its own copy in `test_llm_call_executor.py:105` and is unaffected. Every remaining diff hunk is the same mechanical shape: an inline seven-line `Orchestrator(...)` construction replaced by `_build_service(fake_llm, settings)`, and `orch.` → `service.`. **Not one assertion, `pytest.raises` match, or fixture value changed.** The three cases the ticket names explicitly are intact and untouched: output-language (`:342`, `:368`), hyperlink (`:391`, `:416`), no-trailing-question (`:447`, `:463`).

`_build_service` builds the **real** `LLMCallExecutor` over `FakeLLMPort` / `FakeAnonymizer` — no fake executor — and threads `max_total_tokens` through, so `TestTokenLimit` still asserts a large payload *is* forwarded rather than accidentally asserting nothing.

`tests/api/test_routes.py`, `tests/e2e/` and `tests/integration/` have **zero diff**. The e2e tier can't run here (no Docker/Postgres), but the `integration` CI job is green on this SHA.

### Findings

Five, all non-gating: one minor on the `dependency_overrides` divergence in `tests/api/conftest.py` (the one place this PR's shape differs from the three merged siblings — and it differs because AC 7 forced it to), one minor on a docs paragraph this PR made inaccurate, and three nits. All inline.

One more for the human, with no diff line to hang it on: `docs/operations/settings-reference.md:23-24` still reads "It applies to four call sites: the `analyze` judge, the hierarchical leaf judges, and the judges in `summarize` and `summarize_aggregate`", while the parallel sentence in `03-components.md:59` was updated in this PR to attribute each pair to its owning service. Not wrong — it names methods, and `JUDGE_LLM_*` still routes exactly those four — just the one place the ownership story wasn't carried through. (`summarize_aggregate` for the method now called `summarize_bulk` is a repo-wide inconsistency that predates this PR; the `Operation` enum member really is `SUMMARIZE_AGGREGATE`.)

### Could not verify

- Whether the cross-service judge-routing tests should stay in `tests/services/test_orchestrator_judge_routing.py` or follow the use case into `test_summarize.py` is a taste call. The module docstring argues for keeping them together — "splitting them per service is exactly how one call site quietly stops being checked" — which I find persuasive, but two of the four call sites it covers are no longer on the orchestrator, so the *file name* now undersells its scope. Whether it wants renaming is a human judgement, and renaming it in this PR would contradict the "no renames" instruction.
- The e2e/integration tier was not executed in this environment; I am relying on the green `integration` CI job for it.

mariushelf:
Review iteration 1: **approve** — Fresh review of origin/main...HEAD (merge base 4ee1cce, head fffc997, 20 files, +656/-521). All nine acceptance criteria are met and the branch is green: `make test` = 651 passed / 42 deselected, `make lint` = all checks passed with 3 import-linter contracts kept / 0 broken, and every CI check on this SHA passes (including `integration`, both `test` and both `lint` jobs, and `build-docs`). The iteration-3 blocker (`AnalysisError` dropped from `tests/services/test_orchestrator.py` while a retained `assign_codes` test still used it) is resolved — that file no longer exists on this branch and the `assign_codes` tests live on `main` in `test_coding.py`.

On the human's specific ask — do the four merged PRs and this PR's work both survive, and is anything duplicated: verified directly. `build_services` keeps `main`'s name (the earlier `build_service_graph` rename is gone); `composition.py`'s top-level surface is exactly `main`'s plus one `summarize: SummarizeService` field on `ServiceGraph`. No helper is duplicated: `hyperlink_form_references` has one definition in `qfa.services.record_links` and `JUDGE_USER_MESSAGE` one in `qfa.services.prompts`, both imported by `analyze.py` and `summarize.py`. `summarize.py` imports nothing from `qfa.services.orchestrator` — the private-import coupling flagged in earlier rounds is fully unwound. Naming follows the established convention (`ServiceGraph.summarize` / `app.state.summarize_service` / `get_summarize_service`).

The extraction is faithful: `SummarizeService` is a plain class with `executor: LLMCallExecutor` required (no `None` default), both methods deleted from `Orchestrator`, and a whitespace-normalised diff of both bodies against `origin/main` differs only in the receiver and the removal of the never-read `total_cost` accumulator (dead on `main` too; disclosed in the PR body). Sharing is asserted by identity, not type. The test rename was checked test-by-test: all 8 classes and all 14 functions present, every assertion / `pytest.raises` match / fixture value unchanged, the three ticket-named cases intact, driven over the real `LLMCallExecutor` on `FakeLLMPort`/`FakeAnonymizer`. `tests/api/test_routes.py`, `tests/e2e/` and `tests/integration/` have zero diff.

Five non-gating findings filed inline (2 minor, 3 nit). No blockers or majors.


mariushelf:
Could not verify: Whether the cross-service judge-routing tests should stay in tests/services/test_orchestrator_judge_routing.py or follow the use case into test_summarize.py is a taste call. The module docstring's argument for keeping them together (splitting them per service is how one call site quietly stops being checked) is persuasive, but two of the four call sites it covers are no longer on the Orchestrator, so the file name undersells its scope. Renaming it in this PR would also contradict the reviewer instruction not to rename things.; The e2e/integration tier (42 deselected tests) could not be executed in this environment — no Docker/Postgres available. I relied on the green `integration` CI job on this SHA for it.; docs/operations/settings-reference.md:23-24 still attributes the four judge call sites without service ownership and names `summarize_aggregate` where the method is now `summarize_bulk`, while the parallel sentence in 03-components.md:59 was updated in this PR. It is not wrong (JUDGE_LLM_* does route exactly those four, and `Operation.SUMMARIZE_AGGREGATE` really is the enum member), so whether it is worth updating is a human call. No diff line to anchor an inline comment on; raised in the summary.


olafdegraaff (approved):
LGTC (Looks good to Claude ;) )

PR #281, opened 2026-08-13
Load bundled fixtures via importlib.resources instead of filesystem paths

### LINKED ISSUE #158: Load bundled fixtures via importlib.resources instead of filesystem paths
## Summary

Fixtures and bundled data files are loaded by **constructing filesystem paths relative to the module location** (e.g. `Path(__file__).resolve().parents[3] / "fixtures" / ...`) instead of addressing them as package resources via `importlib.resources`. This is fragile: it assumes the repository directory layout and silently breaks once `qfa` is installed as a (non-editable) wheel, where the repo-root `fixtures/` directory does not exist relative to the installed module.

The canonical example is the coding-framework fixture used to build Swagger examples in `src/qfa/api/schemas.py`.

## Where it happens

**Production code (highest priority — runs at app startup, feeds OpenAPI/Swagger examples):**

- `src/qfa/api/schemas.py:92-124` — `_assign_codes_request_examples()` does:
  ```python
  root = Path(__file__).resolve().parents[3]
  path = root / "fixtures" / "coding_framework.json"
  if not path.is_file():
      return [ ... "no-framework" fallback ... ]
  framework = json.loads(path.read_text(encoding="utf-8"))
  ```
  `parents[3]` walks up out of the installed package to guess the repo root. After a real `pip install` (non-editable) this path lands in `site-packages` and the file is absent, so the endpoint silently degrades to the placeholder "no-framework" example.

**Test code (lower priority — see scope note below):**

- `tests/test_analyze_corpus.py:20-22,45,54` — `Path(__file__).resolve().parent.parent / "fixtures"`
- `tests/test_large_corpus.py:18-19,29` — same pattern
- `tests/scripts/test_stress_analyze.py:23,31,43` — `Path(__file__).resolve().parents[2] / "fixtures" / ...`

## Existing good pattern to follow

`src/qfa/api/composition.py:93-97` already loads bundled data the correct way:

```python
prices_path = importlib.resources.files("qfa.resources").joinpath("model_prices.yaml")
with importlib.resources.as_file(prices_path) as f:
    custom_prices = yaml.safe_load(f.read_text())
```

The fix should make the rest of the codebase consistent with this.

## Proposed change

1. Move data that **production code** needs at runtime into a package resource directory (e.g. `src/qfa/resources/coding_framework.json`, alongside `model_prices.yaml`) and load it via `importlib.resources.files("qfa.resources")`.
2. Ensure `pyproject.toml` packages/ships these resources. Currently `[tool.hatch.build]` only sets `packages = ["src/qfa"]` with no package-data / include for non-`.py` files — confirm `.json`/`.yaml` resources are included in the wheel.
3. Decide on test-only fixtures (see scope note).

## Scope note: runtime data vs. test-only data

Distinguish two cases — they should **not** be solved the same way:

- **Runtime data** (e.g. `coding_framework.json` used by `schemas.py`): belongs inside the package as an `importlib.resources` resource so it ships in the wheel.
- **Test-only fixtures** (e.g. the 2.9 MB `fixtures/analyze_corpus.yaml`, `large_corpus.yaml`): should **not** be bloating the shipped wheel. These can either stay as path-based loads in tests (tests always run from a checkout) or move under a dedicated test package. Avoid blindly bundling large test corpora into the distribution.

## Acceptance criteria

- [ ] `src/qfa/api/schemas.py` loads the coding-framework example via `importlib.resources`, not `Path(__file__).parents[...]`.
- [ ] Swagger "Try it out" examples work from a non-editable install (no silent "no-framework" fallback when the resource is bundled).
- [ ] `pyproject.toml` is configured so bundled resources are included in the build, and this is verified (e.g. inspect the built wheel).
- [ ] Test-only fixtures handling is explicitly decided (kept path-based or moved), with large corpora kept out of the shipped wheel.
- [ ] Docs under `docs/` updated if resource loading / packaging is documented there.

mariushelf:
Committed; tree is clean.

### What changed

**Production fix.** `qfa.api.schemas._assign_codes_request_examples()` built the `POST /v1/assign-codes` Swagger examples by walking `Path(__file__).resolve().parents[3]` to a repo-root `fixtures/` directory. `coding_framework.json` moved to `src/qfa/resources/` (next to `model_prices.yaml`) and is now read with `importlib.resources.files("qfa.resources")`, matching the pattern already used by `register_custom_model_prices()`.

This was not a latent bug — the `Dockerfile` copies `src/` and never `fixtures/`, so the loader's `if not path.is_file()` branch fired in **every deployed environment**, and the OpenAPI schema silently served a single placeholder `"no-framework"` example instead of the 5-root-code taxonomy. The fallback is retained but now logs at WARNING, so a genuinely missing resource is visible rather than silent.

**Single copy of the file.** `tests/test_analyze_corpus.py` and the `--framework` default in `scripts/generate_corpus.py` now read the package resource too (the script default previously only resolved when invoked from the repo root).

**Test-fixture decision (explicit, per the spec's scope note).** `analyze_corpus.yaml` (2.9 MB) and `large_corpus.yaml` (276 KB) stay in repo-root `fixtures/` with path-based loads — tests always run from a checkout, and that data has no business in the wheel. Comments at all three load sites record the choice so it isn't "fixed" later.

**Docs.** New `docs/development/bundled-resources.md` — the runtime-resource vs. test-fixture split, the correct `importlib.resources` / `as_file` idioms, the anti-pattern with the #158 post-mortem, and how to verify wheel inclusion. Linked from `docs/development/index.md` (incl. toctree) and `docs/README.md`. `pyproject.toml` gains a comment on `[tool.hatch.build]` explaining that hatchling ships all tracked files under `packages`, so new resources need no `package-data` entry.

**Tests.** Five new tests in `tests/api/test_schemas.py::TestAssignCodesRequestExamples`: the resource is present as package data; no example carries the `no-framework` marker; every published example validates as an `ApiAssignCodesRequest` (a Swagger "Try it out" that 422s is worse than none); the examples expose the full 3-level taxonomy; and a source guard against reintroducing `parents[...]` arithmetic.

### Verification

- `make test` — 618 passed, 42 deselected.
- `make lint` — ruff, `ty`, and all 3 import-linter contracts pass.
- `make docs` — build succeeded, no new warnings.
- Wheel **and** sdist inspected: both ship `qfa/resources/{__init__.py,coding_framework.json,model_prices.yaml}`; neither contains any `fixtures/` entry.
- The acceptance criterion tested directly in the non-editable shape: extracted the built wheel to `/tmp/wheelsite`, imported it from `cwd=/tmp` with no repo root above it — examples came back as `covid-example-0..2` with 5 root codes. The old code would have fallen back there.
- `scripts/generate_corpus.py`'s default resolves from an arbitrary cwd (154 leaf codes loaded from `/tmp`).

### Impact / risk

Low, and revertible with `git revert`. No API contract, settings, or DB change — the OpenAPI *examples* change from placeholder to real content in deployed environments, which is the intended fix. Only caveat worth flagging: `scripts/generate_corpus.py` now imports `importlib.resources` against an installed `qfa` and wraps `files()` in `Path(str(...))`; that holds for directory-based installs (how this project is always installed) but would not for a zipimported one. It's a manual, non-CI helper script, and the comment says so.

Left undone deliberately: `docs/architecture-review-2026-05-08*.md` still reference the old `fixtures/coding_framework.json` path — those are dated review snapshots, and editing them would falsify the record. Branch is local and unpushed, no PR opened.

Fixes #158


mariushelf:
Build complete. 

mariushelf:

scripts/generate_corpus.py
```diff
@@ -63,6 +65,16 @@
 
 logger = logging.getLogger("generate_corpus")
 
+# Default for ``--framework``: the coding framework ships inside the
+# ``qfa.resources`` package (#158), so this script no longer needs to be
+# invoked from the repo root for the default to resolve. ``files()`` yields a
+# real filesystem path for a directory-based install, which is how ``qfa`` is
+# always installed here; the ``Path(str(...))`` wrap only narrows the
+# ``Traversable`` to the ``type=Path`` the CLI argument expects.
+BUNDLED_FRAMEWORK = Path(
```
mariushelf, line 74: **nit** — this is the one place in the PR that diverges from the rule the new doc lays down. `docs/development/bundled-resources.md` says: if you need a real filesystem path, use `importlib.resources.as_file` and stay inside the context manager. Here the `Traversable` is flattened with `Path(str(...))` at *import time* so it can be an `argparse` default.

The comment is honest about the zipimport caveat and this is a manual, non-CI helper, so it isn't a defect — but it does mean the codebase now demonstrates both the recommended and the discouraged idiom. If you want them to agree, `default=None` plus resolving inside `gen-specs` (`resource.read_text()` when `--framework` was not given) drops the `Path`/`str` round-trip entirely and never needs a real path.

src/qfa/api/schemas.py
```diff
@@ -262,9 +269,14 @@ def _coding_levels_from_framework(framework: dict[str, Any]) -> dict[str, Any]:
...
-    root = Path(__file__).resolve().parents[3]
-    path = root / "fixtures" / "coding_framework.json"
-    if not path.is_file():
+    resource = importlib.resources.files("qfa.resources").joinpath(
+        "coding_framework.json"
+    )
+    if not resource.is_file():
+        logger.warning(
+            "Bundled resource qfa.resources/coding_framework.json is missing; "
+            "serving placeholder assign-codes examples."
```
mariushelf, line 278: **nit** — worth a moment's thought, not a change request: after this PR the fallback is unreachable in every supported install shape (checkout, editable, wheel, and the image — the `Dockerfile` copies `src/`, so the resource is now present there too). A missing `qfa.resources/coding_framework.json` would be a packaging bug, and #158 is precisely the story of a silent degradation being invisible for a long time.

Promoting WARNING to silence-no-more is already the main win, so this is fine as it stands. But raising instead of returning a placeholder would turn the same condition into a startup failure that CI/the image build catches, and would delete ~25 lines of placeholder that nothing tests. Your call — the branch has no test coverage either way.

tests/api/test_schemas.py
```diff
@@ -565,3 +569,53 @@ def test_summarize_bulk_response_output_language_excluded_from_serialization():
...
+    def test_examples_expose_the_real_framework(self) -> None:
+        """The examples carry the full 3-level taxonomy, not a stub."""
+        request = ApiAssignCodesRequest.model_validate(
+            _assign_codes_request_examples()[0]
+        )
+        root_codes = request.coding_levels.root_codes
+        assert len(root_codes) > 1
+        assert max(node.max_child_depth() for node in root_codes) == 2
+
+    def test_loader_does_not_walk_out_of_the_package(self) -> None:
```
mariushelf, line 617: **nit** — a source-text guard is weaker than it looks: it passes if the `parents[...]` arithmetic moves one call away (a module-level constant, a helper, a shared `_repo_root()`), which is the most likely way this actually regresses. The three behavioural tests above it already fail loudly on a real regression, and `test_framework_is_a_package_resource` pins the packaging half.

Harmless to keep — just don't count it as the regression net.

docs/development/bundled-resources.md
```diff
@@ -0,0 +1,99 @@
...
+hatchling ships **every tracked file** under that tree — `.py` or not — so a
+new file dropped into `src/qfa/resources/` needs no `package-data` entry. Do
+verify it after adding one, because a missing resource degrades quietly:
+
+```bash
+uv build --wheel
+unzip -l dist/*.whl | grep qfa/resources
+```
+
+Both the wheel and the sdist are checked; `fixtures/` appears in neither.
```
mariushelf, line 76: **nit** — this sentence records a one-time verification in the present tense, in a doc that nothing re-checks. It was true when written (I rebuilt both artifacts and confirmed: wheel and sdist ship `qfa/resources/{__init__.py,coding_framework.json,model_prices.yaml}`, and neither contains a `fixtures/` entry), but a future `force-include` or a stray large file under `src/qfa/` would silently falsify it.

Either phrase it as an instruction ("check both the wheel and the sdist") or, better, let the `unzip -l` snippet above be the whole story.

mariushelf (commented):
## Review — #158, fresh context

**Verdict: approve.** The diff does what the ticket asked, and the one thing the ticket asked to be *verified* (wheel inclusion) I re-verified from scratch rather than taking the summary's word for it.

### On the loop's "1 of 9 tasks committed" note

False alarm. `tasks/plan.md` lists nine steps; the single commit covers all nine — the move (1), the `schemas.py` rewrite incl. dropping the `Path` import (2), `tests/test_analyze_corpus.py` (3), the `generate_corpus.py` default (4), `pyproject.toml` (5), new tests (6), docs (7) — and steps 8–9 are verify-and-commit, which produce no separate diff. Nothing is missing.

### Independent verification

I re-ran everything rather than trusting the PR body:

- `make test` — 618 passed, 42 deselected.
- `make lint` — ruff, `ty`, 3/3 import-linter contracts kept.
- `make docs` — build succeeded.
- `uv build --wheel` → the wheel ships `qfa/resources/{__init__.py,coding_framework.json,model_prices.yaml}` and contains no `fixtures/` entry. Same for the sdist.
- **The acceptance criterion that matters**, tested in the non-editable shape: extracted the wheel to a bare directory, imported `qfa` from `cwd=/tmp` with no repo root anywhere above it, called `_assign_codes_request_examples()` → `['covid-example-0', 'covid-example-1', 'covid-example-2']`, 5 root codes. `parents[3]` from there would have landed in the extract dir's parent and hit the placeholder.
- The PR's claim about production impact holds by inspection: `Dockerfile` copies `pyproject.toml`, `src/`, `alembic/`, `entrypoint.sh` — never `fixtures/` — so the `is_file()` fallback did fire in the image, and now doesn't.
- No live consumer of the old path is left behind: the only remaining `fixtures/coding_framework.json` string is in `docs/architecture-review-2026-05-08-addendum.md`, a dated review snapshot the author deliberately left intact. That's the right call — a review record that's been edited to match later code isn't a record.

### Acceptance criteria

| Criterion | |
|---|---|
| `schemas.py` loads via `importlib.resources`, not `parents[...]` | met |
| Swagger examples work from a non-editable install | met — verified directly, see above |
| `pyproject.toml` ships the resources, verified against the build | met — wheel + sdist inspected |
| Test-fixture handling explicitly decided, corpora out of the wheel | met — decision recorded at all three load sites and in the docs; 2.9 MB + 276 KB stay in repo-root `fixtures/` |
| Docs updated | met — new `development/bundled-resources.md`, linked from both indexes and the toctree |

### What I liked

The `pyproject.toml` change is a comment rather than a `[tool.hatch.build.targets.wheel]` block, and that's the better choice: hatchling's `packages` already ships every tracked file under the tree, so an explicit `artifacts`/`force-include` entry would be config that looks load-bearing while doing nothing — and would need editing every time a resource is added. The comment plus the `unzip -l` recipe is the honest version.

Making the tests and `generate_corpus.py` read the same package resource — rather than leaving a second copy in `fixtures/` — is what keeps this fix from decaying.

### Findings

Four inline, all nits, none gating: the `Path(str(...))` in `generate_corpus.py` diverging from the `as_file` idiom the new doc prescribes; the now-unreachable and untested placeholder fallback (raise vs. degrade is a judgment call I'm leaving to you); the `inspect.getsource` regression guard being weaker than the behavioural tests beside it; and a docs sentence stating a one-time artifact check in the present tense.

mariushelf:
Review iteration 1: **approve** — The PR does what #158 asked. `coding_framework.json` moved from repo-root `fixtures/` into `src/qfa/resources/` (next to `model_prices.yaml`) and `qfa.api.schemas._assign_codes_request_examples()` now reads it via `importlib.resources.files("qfa.resources")` instead of `Path(__file__).resolve().parents[3]`. Tests and `scripts/generate_corpus.py` were pointed at the same package resource so only one copy of the file exists. The test-only corpora (`analyze_corpus.yaml` 2.9 MB, `large_corpus.yaml` 276 KB) deliberately stay path-based in repo-root `fixtures/`, with the decision recorded in a comment at each of the three load sites and in a new docs page.

I re-verified independently rather than trusting the PR body: `make test` (618 passed, 42 deselected), `make lint` (ruff, ty, 3/3 import-linter contracts kept), `make docs` (build succeeded). `uv build --wheel` and `--sdist` both ship `qfa/resources/{__init__.py,coding_framework.json,model_prices.yaml}` and neither contains any `fixtures/` entry. Most importantly I exercised the headline acceptance criterion in the non-editable shape: extracted the wheel to a bare directory and imported `qfa` from `cwd=/tmp` with no repo root above it — `_assign_codes_request_examples()` returned `covid-example-0..2` with 5 root codes, where the old `parents[3]` walk would have hit the placeholder. The PR's production-impact claim also holds by inspection: the Dockerfile copies `src/` and never `fixtures/`, so the silent `no-framework` fallback did fire in every deployed image.

The loop's "1 of 9 tasks committed" note is a false alarm — the single commit covers all nine plan steps; steps 8–9 are verify-and-commit, which produce no separate diff.

Four findings, all nits, none gating: the `Path(str(files(...)))` flattening in `generate_corpus.py` diverges from the `as_file` idiom the PR's own new doc prescribes; the placeholder fallback in `schemas.py` is now unreachable in every supported install shape and untested (raise-vs-degrade is a maintainer judgment call); the `inspect.getsource` regression guard is weaker than the behavioural tests beside it; and a docs sentence records a one-time artifact check in the present tense with nothing re-checking it.


mariushelf:
**Review verdict: approved (tier 2)**
Could not verify: Whether the rendered Swagger UI in a live deployment displays the examples correctly end to end — I verified the example payloads are produced from an unpacked wheel and validate as `ApiAssignCodesRequest`, but did not build/run the container image or exercise "Try it out" against a running server.; Whether shipping the 128 KB `coding_framework.json` inside the wheel (parsed at `qfa.api.schemas` import time on every process start) is acceptable to the maintainers — this is a product/packaging preference, not something the diff can settle. It was already parsed at import time from a checkout, so nothing regressed.; Whether leaving `docs/architecture-review-2026-05-08-addendum.md:738`'s reference to the old `fixtures/coding_framework.json` path is the right call. The author's reasoning (dated review snapshots shouldn't be retro-edited) is sound and I agree, but whether those snapshots are treated as immutable records in this repo is a human convention I can't confirm from the diff.


olafdegraaff:
LGTM

olafdegraaff (approved):

PR #283, opened 2026-08-14
docs+test: address review findings on #279 (CodingService extraction)

### LINKED ISSUE #265: refactor(services): extract CodingService from Orchestrator

mariushelf:
Follow-up to #279, which merged while the review findings were being addressed. Same findings, applied on top of `main`.

## What changed

**Lost-in-rebase content from #275 restored** — this is the substantive one. #279 was rebased onto `main` after #275 landed, and the rebase silently reverted #275's rewritten `assign_codes` description in `docs/architecture/03-components.md`, putting the superseded *"Multiple LLM calls per record: pick + judge at each level"* text back. Restored to #275's wording (*"One LLM call picks the best-fitting code path(s) directly from the whole flattened coding framework, then a separate judge call per level scores each selected path, root to leaf."*).

I checked the rest of #275 the same way: its changes to `llm_client.py`, `domain/errors.py`, `coding_classifier.py`, `rest-api/index.md` and `routes.py` are all intact on `main`, all 17 tests it added are present, and the code moved into `coding.py` is byte-identical to `main`'s post-#275 `orchestrator.py`. The one doc row was the only casualty.

**Also corrected while in that file:** `docs/architecture/01-architecture-style.md` claimed `detect_sensitive_content` still sits on `Orchestrator`. #277 extracted it; the method is gone from the class.

**Two service tables folded into one** (`03-components.md`) — `CodingService` and `SensitivityService` appeared both in "The application services" and in "Extracted use-case services", two rows to keep in step per extraction with more of epic #112 coming. The provider column moves into the first table and the second is dropped; when `Orchestrator` disappears with #267 the remaining table is already the right one. The `#the-application-services` anchor is unchanged, so the link from `implementing-a-new-endpoint.md` still resolves.

**`tests/api/conftest.py`** — the fixture binds one `FakeOrchestrator` to three `app.state` slots at setup time, so the `TestErrorMapping` idiom of rebinding `state.orchestrator` does *not* reach `/v1/detect-sensitive` or `/v1/assign-codes`. Nothing regresses today — no assign-codes error-mapping test exists — but one written to the house idiom would pass vacuously against the healthy double. The comment now says that plainly and points at the slot to set instead.

**`tests/services/test_orchestrator_judge_routing.py`** — in `test_coding_classification_stays_entirely_on_the_primary`, the `RoutingLLM("judge")` local was handed to nobody once `CodingService` stopped taking a judge client, so `assert judge.calls == []` held regardless of what the code did. Dropped the local and the assertion; the docstring now claims only the behavioural half the test still checks (`primary.response_models == [CodingResponse, JudgeResponse]`) and notes that the structural half is enforced by the constructor signature, not here.

**Coding module docstring** widened: post-#275 `CodingService` uses the executor for the deadline-derived timeout as well as the token-budget guard.

## Findings deliberately not acted on

Three review findings said so themselves and I agree:
- `composition.py` — `build_orchestrator` building the full `ServiceGraph` and discarding the `CodingService`: explicitly "no change requested".
- `tests/services/test_coding.py` importing doubles from `test_orchestrator`: worth moving to `tests/services/conftest.py` when a third module wants the same pair, and it would mean editing a test file no finding names.
- `coding.py` — the two `self._llm.complete(...)` calls without an explicit `timeout=`: pre-existing and byte-identical to `main`, out of scope for a pure extraction. Worth a follow-up in the epic that routes them through `executor.complete()`.

## Verification

`make lint` — ruff, `ty`, all three import-linter contracts kept. `make test` — 644 passed. `make docs` builds clean.

Not run: the 42 `integration`/`e2e` tests are deselected by default and need Postgres, unavailable in this container. No source behaviour changed here, only docs, comments and one dead assertion.

Follow-up to #279 · Fixes #265

### DIFF
--- a/docs/architecture/01-architecture-style.md
+++ b/docs/architecture/01-architecture-style.md
@@ -35,7 +35,7 @@ Allowed import directions (enforced by `import-linter`):
 
 ## What's *not* hexagonal here
 
-Hexagonal tells us "services depend only on ports" — it doesn't say "one orchestrator class with N methods" versus "N orchestrator classes." [ADR-011](../adr/011-drop-orchestrator-port.md) started from one {py:class}`~qfa.services.orchestrator.Orchestrator` holding every operation and anticipated extracting individual use cases once one grew enough to warrant it; [ADR-017](../adr/017-orchestrator-composition-only.md) takes that escape valve and splits the class into one service per use case, sharing behaviour by composition rather than a base class. {py:class}`~qfa.services.coding.CodingService` (assign-codes) is extracted; `analyze`, `summarize` and `detect_sensitive_content` still sit on `Orchestrator` while the epic runs.
+Hexagonal tells us "services depend only on ports" — it doesn't say "one orchestrator class with N methods" versus "N orchestrator classes." [ADR-011](../adr/011-drop-orchestrator-port.md) started from one {py:class}`~qfa.services.orchestrator.Orchestrator` holding every operation and anticipated extracting individual use cases once one grew enough to warrant it; [ADR-017](../adr/017-orchestrator-composition-only.md) takes that escape valve and splits the class into one service per use case, sharing behaviour by composition rather than a base class. {py:class}`~qfa.services.sensitivity.SensitivityService` (detect-sensitive) and {py:class}`~qfa.services.coding.CodingService` (assign-codes) are extracted; `analyze` and `summarize` still sit on `Orchestrator` while the epic runs.
 
 ## Further reading
 
--- a/docs/architecture/03-components.md
+++ b/docs/architecture/03-components.md
@@ -101,34 +101,30 @@ the rationale, and flow/sequence diagrams.
 
 Each use case is one async method backing one HTTP endpoint:
 
-| Service | Method | Endpoint | What it does |
-|---|---|---|---|
-| {py:class}`~qfa.services.orchestrator.Orchestrator` | `analyze` | `POST /v1/analyze` (`mode=single_pass`) | One LLM call. Free-text summary of themes across submitted records. |
-| {py:class}`~qfa.services.orchestrator.Orchestrator` | `analyze_hierarchical` | `POST /v1/analyze` (`mode=hierarchical`) | Embed -> cluster -> map -> reduce pipeline. Returns additional `confidence` and `coding_trends` fields. |
-| {py:class}`~qfa.services.orchestrator.Orchestrator` | `summarize` | `POST /v1/summarize` | One LLM call. Per-record summaries with a self-evaluated quality score. |
-| {py:class}`~qfa.services.coding.CodingService` | `assign_codes` | `POST /v1/assign-codes` | Multiple LLM calls per record: pick + judge at each level of a hierarchical coding framework. |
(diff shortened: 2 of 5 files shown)


mariushelf:
Issue #265 was closed while this PR was being built, so the loop cannot take it any further: a closed ticket has no next lifecycle state, and the loop will not merge or close finished work on its own.

The build completed and this PR is intact. Handing it to a human to merge it or close it deliberately.

